PR #12, opened 2020-07-22
Start of specification implementation

bram2000:


digital_land/specification.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, path):
+        self.dataset = {}
+        self.dataset_names = []
+        self.schema = {}
+        self.schema_names = []
+
+        self.load_datasets(path)
+        self.load_schemas(path)
+
+    def load_datasets(self, path):
```
psd, line 15: Why not keep these singular?

bram2000: Fair point - will change.

PR #16, opened 2020-09-24
Do not blow up if mapping and empty file

bram2000:


bram2000:
Not sure if we still need this. I expect the recent collection fixes will address the empty files but will leave this open for now.

psd:
Should we delete this PR @bram2000 ?

PR #22, opened 2020-10-21
Collection resource

psd:
Collection is a dataset which can be loaded from a collection directory, or in the future a data-package.

```
from digital_land.collection import Collection

collection = Collection()
collection.load()

for resource in collection.resources():
    print(resource, collection.resource_organisation(resource))

collection.log.save("tmp/log.csv")
```

### DIFF
--- a/digital_land/cli.py
+++ b/digital_land/cli.py
@@ -62,12 +62,6 @@ def issue_path(f):
     )(f)
 
 
-def index_dir(f):
-    return click.option(
-        "--index-dir", "-x", type=click.Path(exists=True), default="index/"
-    )(f)
-
-
 @click.group()
 @click.option("-d", "--debug/--no-debug", default=False)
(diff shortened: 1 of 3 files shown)


digital_land/cli.py
```diff
@@ -96,6 +96,31 @@ def collect_cmd(endpoint_path):
...
+    # TBD: replace with Collection()
+    indexer = Indexer()
+    indexer.index()
+
+
+@cli.command("collection-list-resources", short_help="list resources for a pipeline")
+def pipeline_resources_cmd():
+    collection = Collection()
+    # can be loaded from a collection directory, or a collection datapackage
+    collection.load()
```
bram2000, line 116: Can we call `load` from the constructor of `Collection`? Feels like it's part of initialising the object.

ksp37 (approved):

PR #26, opened 2020-10-28
Command to check failing endpoints

ksp37:


digital_land/update.py
```diff
@@ -0,0 +1,98 @@
...
+        return False, failure_reason
+
+    if "resource" not in log_item:
+        failure_reason = "Resource not retrieved"
+        return False, failure_reason
+
+    return True, failure_reason
+
+
+def get_entries_between_keys(start_key, end_key, length, register_lookup):
```
ksp37, line 69: This is generic enough to be used on any sorted register....so could move to registers.py

bram2000 (changes requested):
Looks really good Kishan - just change to write stdout and I think we're golden :)

bram2000 (approved):

PR #27, opened 2020-11-03
Implement command to add sources and endpoints

ksp37:


digital_land/update.py
```diff
@@ -66,33 +70,151 @@ def has_collected_resource(log_item):
...
+                    print(
+                        "Active source entry already exists for organisation {} and URL {}".format(
+                            resource_entry["organisation"],
+                            resource_entry["endpoint-url"],
+                        )
+                    )
+                    return
+
+    source_entry = {
+        "collection": resource_entry.get("pipeline", ""),
```
bram2000, line 165: you could use a `defaultdict(str)` here in place of the dict for resource_entry, it would avoid the use of `.get` all over the place, because `""` would be the default for any missing key.

digital_land/update.py
```diff
@@ -66,33 +70,151 @@ def has_collected_resource(log_item):
...
+    if "start-date" in resource_entry:
+        source_entry["start-date"] = (
+            datetime.strptime(resource_entry["start-date"], "%Y-%m-%d").date()
+            if resource_entry["start-date"]
+            else ""
+        )
+    else:
+        source_entry["start-date"] = date.today().strftime("%Y-%m-%d")
+
+    source_item = Item(source_entry)
```
bram2000, line 188: does `Item(source_entry)` _have_ to contain all of the expected keys for a register? or can it be sparsely populated and the register fall back to empty values/defaults?

It feels like a register should allow some empty values, in which case maybe we can avoid explicitly setting all of the optional values for e.g. attribution, licence, end-date etc.

digital_land/cli.py
```diff
@@ -65,6 +69,15 @@ def issue_path(f):
     )(f)
 
 
+def source_endpoint_dir(f):
+    return click.option(
+        "--source-endpoint-dir",
```
bram2000, line 74: I _think_ this would be better specifying the two files separately as `endpoint-path` and `source-path`. We already use `endpoint-path` for the collect action so can share the option definition with that. Willing to have my mind changed though!

bram2000 (changes requested):

digital_land/cli.py
ksp37: I debated myself with the same question (I was wondering whether the two files would ever live in different directories). But two paths is more flexible anyway, I'll do that 👍

digital_land/update.py
ksp37: So, again, I thought similar - just iterating over `resource_entry` and creating a dict from the existing keys (it would be neater). But I was torn because some entries, e.g. collection, entry-date, end-date have particular handling, so I thought I may as well just make everything explicit. But I can iterate to create the dict and then handle the special cases if that's better?

bram2000: that's fine, explicit is better and I don't think we'd save much. Maybe just do the defaultdict change and that should neaten it up anyway.

digital_land/update.py
ksp37: Updated

PR #50, opened 2020-12-01
feat: pluggy

bram2000:


digital_land/cli.py
```diff
@@ -310,7 +316,7 @@ def pipeline_cmd(input_path, output_path, null_path, issue_path, use_patch_callb
         collection,
         Organisation().organisation_uri,
         patch,
-        use_patch_callback,
+        pm,
     )
```
ksp37, line 320: No `pm.register(harmoniser)` needed? (Like L255)

digital_land/harmonise.py
```diff
@@ -153,12 +152,13 @@ def harmonise(self, reader):
...
-                if self.harmonise_callback and hasattr(
-                    self.harmonise_callback, "patch"
-                ):
-                    row[field] = self.harmonise_callback.patch(
-                        field, row[field], self.log_issue
+                if self.plugin_manager:
+                    plugin_results = self.plugin_manager.hook.apply_patch_post(
+                        fieldname=field, value=row[field]
                     )
+                    if len(plugin_results) == 1:
```
ksp37, line 159: It can return an array?

bram2000: yes, `plugin_manager.hook.<function>` will always return a list, as you might have multiple implementations for a given hook.

digital_land/cli.py
bram2000: good spot, this may actually be better in `__init__` of harmoniser, so it always happens. Will change now.

ksp37 (approved):

PR #61, opened 2021-02-22
Method to return parent of field

colmjude:


digital_land/specification.py
```diff
@@ -147,6 +147,12 @@ def field_type(self, fieldname):
 
         raise ValueError("unknown datatype '%s' for '%s' field" % (datatype, fieldname))
 
+    def field_parent(self, fieldname):
+        field = self.field[fieldname]
+        if field["parent-field"]:
+            return field["parent-field"]
+        return ""
```
bram2000, line 154: I think I'd tend to return `None` in this case rather than empty string. Certainly not a rule but just feels better to mean "no result".

tests/unit/test_specification.py
```diff
@@ -103,6 +103,11 @@ def test_field_typology():
     assert specification.field_typology("field-categories") == "category"
 
 
+def test_field_parent():
```
bram2000, line 106: might as well add a test for a missing parent-field returning "" or None (depending on outcome of previous comment).

bram2000 (changes requested):

digital_land/specification.py
colmjude: I'd normally go with `None` too but noticed `""` had been used in the `field_typology` method [line 153](https://github.com/digital-land/digital-land-python/blob/main/digital_land/specification.py#L153)

colmjude: Sorted this with 43cfbd597

tests/unit/test_specification.py
colmjude: If the specification csv is loaded the `parent_field` will always exist so we decided a test for returning None isn't necessary.

digital_land/specification.py
colmjude: And 8cbf333

bram2000 (approved):

PR #62, opened 2021-03-15
E2e test

ksp37:


tests/e2e/test_workflow.py
```diff
@@ -0,0 +1,144 @@
...
+import filecmp
+from pathlib import Path
+
+from tests.utils.helpers import hash_digest, execute, print_diffs
+
+ENDPOINT = "https://raw.githubusercontent.com/digital-land/digital-land-python/e2e_test/tests/data/resource_examples/e2e.csv"
+
+
+@pytest.fixture()
+def create_workspace(tmp_path):
```
bram2000, line 12: normally fixtures are named in a declarative (?) rather than imperative style, so instead of `create_workspace` which is a command, you could just use `workspace` or something. I read it in my head as "test_workflow is passed a workspace".

tests/integration/test_collect.py
```diff
@@ -1,20 +1,16 @@
...
 import pytest
 
-ENDPOINT = "https://www.registers.service.gov.uk/registers/country/download-csv"
-COLLECTION_DIR = "./collection"
+ENDPOINT = "https://raw.githubusercontent.com/digital-land/digital-land-python/main/tests/data/resource_examples/csv.csv"
 
 
 @pytest.fixture()
-def endpoint_csv(tmp_path):
+def create_endpoint_csv(tmp_path):
```
bram2000, line 13: again avoid imperative language in fixture names. "test_collect is passed an endpoint_csv"

tests/integration/test_collect.py
```diff
@@ -24,11 +20,15 @@ def endpoint_csv(tmp_path):
...
+@pytest.fixture()
+def create_collection_dir(tmp_path):
+    c = tmp_path / "collection"
+    c.mkdir()
+    return c
 
 
-def test_collect(endpoint_csv):
+def test_collect(create_endpoint_csv, create_collection_dir):
+    collection_dir = create_collection_dir
```
bram2000, line 31: this line is avoided if you remove the `create_` prefix.

tests/integration/test_collect.py
```diff
@@ -37,24 +37,21 @@ def test_collect(endpoint_csv):
             "-s",
             "tests/data/specification",
             "collect",
-            endpoint_csv,
+            "-c",
+            collection_dir,
+            create_endpoint_csv,
```
bram2000, line 42: and this reads oddly because of the imperative name.

tests/unit/test_specification.py
```diff
@@ -67,6 +71,12 @@ def test_field_names():
         "field-old",
         "field-category",
         "field-categories",
+        "name",
```
bram2000, line 74: this is what I meant about having to add more noise to the existing unit tests.

tests/utils/helpers.py
```diff
@@ -0,0 +1,47 @@
+import subprocess
```
bram2000, line 1: maybe things like this could go into a `conftest.py` file, which is automatically sources and available for use inside the tests.

bram2000 (commented):
Good stuff. Some stylistic comments added.

PR #63, opened 2021-03-17
feat: entry model, repo and loader

bram2000:


digital_land/cli.py
```diff
@@ -291,6 +293,24 @@ def transform_cmd(input_path, output_path):
...
+        print("missing output path")
+        sys.exit(2)
+
+    repo = EntryRepository(output_path, create=True)
+    loader = EntryLoader(repo)
+
+    total = len(input_paths)
+    for idx, path in enumerate(input_paths, start=1):
+        logging.info("loading file %s of %s", idx, total)
+        stream = load_csv_dict(path, line_num=True)
```
ksp37, line 310: `line_num=True` --> `include_line_num=True`

digital_land/load.py
```diff
@@ -32,23 +32,32 @@ def resource_hash_from(path):
 
 
 class DictReaderInjectResource(csv.DictReader):
-    def __init__(self, resource, *args, **kwargs):
+    def __init__(self, resource, include_line_num, *args, **kwargs):
         self.resource = resource
+        self.include_line_num = include_line_num
+        if include_line_num:
+            self.line_num = 1
```
ksp37, line 39: Does this need updating after the convo on Slack ?

digital_land/repository/entry_repository.py
```diff
@@ -0,0 +1,198 @@
...
+        )
+        cursor.execute(
+            """CREATE TABLE provenance (
+                entry INTEGER,
+                fact INTEGER,
+                entry_date,
+                start_date,
+                end_date,
+                FOREIGN KEY(entry) REFERENCES entry(id),
+                FOREIGN KEY(fact) REFERENCES fact(id)
```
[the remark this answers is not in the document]
bram2000: Yeah maybe; it won't currently make any difference as we always throw the whole thing away and rebuild it, and the nature of it means that it will never clash on provenance. If and when we start updating things in place, this will be more important.

digital_land/repository/entry_repository.py
```diff
@@ -0,0 +1,198 @@
...
+        )
+        cursor.execute(
+            """SELECT rowid FROM fact WHERE entity = ? AND attribute = ? AND value = ?""",
+            (fact.entity, fact.attribute, fact.value),
+        )
+        return cursor.fetchone()[0]
+
+    def _insert_provenance(self, cursor, entry_id, fact_id):
+        cursor.execute(
+            """INSERT INTO provenance(entry, fact) VALUES(?,?)""",
```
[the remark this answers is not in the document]
bram2000: Same reason as above - if we get clashes here it means something is wrong. I'll add the constraint in at the provenance table level but leave this as is - if it hits the constraint we should figure out what's wrong.

digital_land/cli.py
bram2000: good spot - I changed the name half way through development. I wonder why this doesn't break!

ksp37: It did haha. I noticed it when trying to run it locally

digital_land/load.py
bram2000: no this starts counting from 1 (ignoring header)

digital_land/cli.py
bram2000: ah damn. Pretty much the only bit of code not covered by a unit test. This is the kind of plumbing that an e2e test would catch. Or we'd have to mock the EntryLoader and check it's called properly from cli.py.

ksp37 (approved):

PR #65, opened 2021-03-19
feat: store entry-dates and support ordering

bram2000:


digital_land/model/entry.py
```diff
@@ -17,17 +18,28 @@ def __init__(self, data: Item, resource: str, line_num: int = None):
...
         return hash((self.resource, self.line_num, frozenset(self.data.items())))
 
     @classmethod
     def from_facts(
-        cls, entity: str, facts: Set[Fact], resource: str, line_num: int = None
+        cls,
+        entity: str,
+        facts: Set[Fact],
+        resource: str,
+        line_num: int,
```
ksp37, line 35: There's no longer a default for `line_num` - was that intentional?

digital_land/repository/entry_repository.py
```diff
@@ -188,12 +188,13 @@ def _entries_from_rows(self, entity, rows):
...
             line_num = rowdict.pop("__LINE_NUM__")
-            entry_fact[(resource, line_num)].add(
+            entry_date = rowdict.pop("__ENTRY_DATE__")
+            entry_fact[(resource, line_num, entry_date)].add(
                 Fact(entity, rowdict["attribute"], rowdict["value"])
             )
 
         result = {
-            Entry.from_facts(entity, facts, key[0], key[1])
+            Entry.from_facts(entity, facts, key[0], key[1], key[2])
```
ksp37, line 197: Could use argument expansion (not sure if it will work), i.e. 
`Entry.from_facts(entity, facts, *key)`

digital_land/repository/entry_repository.py
```diff
@@ -124,9 +127,6 @@ def _create_schema(self):
                 entity TEXT,
                 attribute TEXT,
                 value TEXT,
-                entry_date,
-                start_date,
```
ksp37, line 128: Why were these removed?

ksp37 (approved):

digital_land/model/entry.py
bram2000: yes - we shouldn't see missing line numbers unless there's a bug.

digital_land/repository/entry_repository.py
bram2000: nice idea - will try.

digital_land/repository/entry_repository.py
bram2000: we currently don't have a meaning for them. Can add them again easily if required.

PR #89, opened 2022-04-07
Simplify organisation lookups -- DO NOT MERGE UNTIL PASSING TESTS

psd:
This is quite a large refactoring of the harmonise phase, and removal of the plugin architecture which isn't needed as we centralise our processing.

At this point the integration tests are failing because the pipeline now provides the organisation-entity, not the assembling of the dataset, and the test data needs updating to reflect that change.


psd:
Updated the branch. We need to remove the listed-building data and construct better acceptance tests as that collection is being restructured and their size makes it almost impossible to spot why they're breaking.

PR #106, opened 2022-11-08
Feat/change fact load

eveleighoj:
What?
add function and change how the fact table is uploaded form the transformed csv files.

Why?
Replace was being use don conflict, this meant that the facts present in the fact table were not the most recent version of the facts due to their being no ordering within the files. The new function is only relevant for the fact table.

### DIFF
--- a/digital_land/package/dataset.py
+++ b/digital_land/package/dataset.py
@@ -203,6 +203,7 @@ class DatasetPackage(SqlitePackage):
                 colname(table),
                 ",".join([colname(field) for field in fields]),
                 ",".join(["%s" % self.colvalue(row, field) for field in fields]),
+                ",".join([colname(field) for field in conflict_fields])
                 ", ".join(["%s=excluded.%s" % (colname(field),colname(field)) for field in update_fields]),
                 colname(table)
             )
--- a/digital_land/package/dataset.py
+++ b/digital_land/package/dataset.py
@@ -4,7 +4,7 @@ import re
 import shapely.wkt
 from decimal import Decimal
 import logging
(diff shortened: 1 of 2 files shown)


digital_land/package/dataset.py
```diff
@@ -185,14 +185,48 @@ def add_counts(self):
         self.commit()
         self.disconnect()
 
+    def entry_date_upsert(self, table, fields, row, conflict_fields, update_fields):
+        """
+        Dataset specific upsert function that only replace values for more recent entry_dates.
+        Will insert rows where no conflict is found. where there's a conflict it was compare entry dates
```
ashimali, line 191: It will insert rows where no conflict is found. If there's a conflict it will compare entry dates and 
update to the more recent.

ashimali:
It looks like we also already have a test called test_package_dataset.py.

Is it possible to merge these (I haven't looked in detail at what test_package_dataset.py does) and remove the package sub directory test package that test_dataset is in?  I'm not sure we need another directory (package), so my vote would be flatter. i.e. test -> integration -> load of tests.

However code itself looks reasonable and good to go apart from my pedantic comment on your doc string.

ashimali:
Oh and yes run lint and black.

eveleighoj:
I'm keen on structuring the unit tests and integration tests on the code repository itself i.e. all tests that are testing the classes and methods in the package > dataset.py should be in the files package > test_dataset.py. Happy to be overruled but it should help when adding new tests as it's easy to find the tests that already exist etc.

ashimali:
👍🏼 merge away once black and lint

PR #112, opened 2023-01-25
Sniff coordinates in a MultiLineString

psd:
Fail more gracefully on Newcastle's article 4 directions data which contains a MultiLineString.

The code now flags geometries beyond point, line and polygons as being invalid when testing coordinates, but I think they would similarly raise an exception with the old code anyway.


eveleighoj (approved):

psd:
Checked with @eveleighoj

PR #121, opened 2023-04-25
restricting loading of old_entity table

ssadhu-sl:


eveleighoj:
Need to increase the testing prior to including, I will completely remove the execute many statement and look at adding a test on the dataset side that ensures that load old entity function only uses

eveleighoj (approved):

PR #122, opened 2023-05-05
Feat/column mapping

morgan-sl:
Currently, column-mapping in the processing pipeline is performed using 'resource' hash codes.
These changes add the ability to use endpoints, as well as resource, hashes for column-mapping.

This initial commit adds the latest specification files from the specification repo.
The content of the repo files better represent the 'real world' data required by the integration test.
The directory can be swapped out to point to an 'as-live' folder, as and when required.


digital_land/pipeline.py
eveleighoj, line 1: When your combining resource mappings and endpoint mappings what happens when both mappings have the same keys? Will it error out?

digital_land/pipeline.py
```diff
@@ -210,6 +227,8 @@ def default_fields(self, resource=None):
         return d
 
     def default_values(self, endpoints=[]):
+        if endpoints is None:
```
eveleighoj, line 230: Is this just to account for when endpoints is accidentally set to None?

.gitignore
```diff
@@ -23,4 +23,7 @@ var
 .junitxml
 pyrightconfig.json
 .idea
-.python-version
\ No newline at end of file
+.python-version
+tests/data/collection-*
```
eveleighoj, line 27: This seems odd when are we excluding folders from tests

tests/integration/column_mapping/helper/funcs.py
eveleighoj, line 1: This seems like a large amount of data is there a reason why you can't use 2 or three rows?

tests/integration/column_mapping/helper/funcs.py
```diff
@@ -0,0 +1,265 @@
...
+    columns_data.to_csv(data_src, index=False)
+
+
+def process_pipeline(params=None):
+    if params is None:
+        params = {}
+    resource = params["resource"]
+    pipeline = params["pipeline"]
+    endpoints = params["endpoints"]
+    specification = params["specification"]
```
eveleighoj, line 240: is the specification needed for this step?

tests/integration/column_mapping/specification/attribution.csv
eveleighoj, line 1: I'm not convinced we need to store another copy of the specification here

tests/integration/column_mapping/test_pipeline_column_mapping.py
```diff
@@ -0,0 +1,237 @@
...
+    output_dir = tmp_path / "output"
+    output_dir.mkdir()
+
+    yield
+
+    # Teardown:
+    pass
+
+
+def test_required_external_asset_creation(env_vars):
```
eveleighoj, line 55: We don't need to test that the specification is being read in properly here. That would be done in the specification testing.

tests/integration/column_mapping/test_pipeline_column_mapping.py
eveleighoj, line 1: Unit and integration tests should mimick the python directory structure based on the functions/classes/methods you are running. If you are performing an integration on the Map Phase it should be in tests > integration > phase > map.py . if we don't stick to a clear structure then the directory will become chaotic and unmaintainable. e2e or acceptance tests should be organised based on use of the product. unit and integration should be function based

eveleighoj: There are also a huge amount of nested fixtures which are incredibly specific to this set of tests. I would be hard for someone else to pick these up and use them in their own tests.

digital_land/pipeline.py
morgan-sl: As I understand it, they are hashes. Hashes are supposed to be resistant to this so, if we find this happening a lot, we might wish to consider changing / upgrading the hash algorithm used rather than trying to code against it.

.gitignore
morgan-sl: My mistake.
I didn't intend to check this i

tests/integration/column_mapping/helper/funcs.py
morgan-sl: Fair point.
I've reduced the datasets down to one example each of:
- standard field
- resource-specific column mapping
- endpoint-specific column mapping

tests/integration/column_mapping/specification/attribution.csv
morgan-sl: Agreed.
Was something I was hoping to discuss with you.
The specification folder in tests/data is not as up-to-date as the specification repo.
I wanted to ensure that my tests were running against the latest changes - as will be the case in PROD.
I wondered whether you'd be able to point it to the latest version, perhaps by changing the env_vars directory.
I also didn't want to replace what is already in data/specification because that'd probably break a lot of existing tests.
Happy to discuss further.

tests/integration/column_mapping/test_pipeline_column_mapping.py
morgan-sl: Accepted.
Happy to remove the test if it's redundant.

digital_land/pipeline.py
eveleighoj: resources and endpoints are hashes, resource_columns and endpoint_columns in the code below I believe are dictionaries of column mappings. These can overlap e.g. is endpoint columns is 
{'REF':'reference'}
and resource_columns are:
{'REF':'name'}
Then what will happen? will it error out?

PR #126, opened 2023-07-20
Feat/add collection source

morgan-sl:
Core functionality ready for review and additional testing.


eveleighoj:
I've started having a look through your code, I'm going to come up with a rough plan on how I think we should take this forward. While I think this looks good as a start we should be aiming to improve the code base as a whole rather than creating solely new functions for specific processes. 

this should include expanding and improving methods within the classes that we have to make there more malleable for different purposes through refactoring. Much of the code may either be available or should be available through the classes that you partially use.

For example there is an organisation class that is generally used to interact with the organisation dataset. this could have methods to download the organisation from the default source. 

Also much of the preprocessing that you do to clean up previous runs may not be necessary here as it already should be present when interacting with this code. We can investigate how to keep some of this functionality incase it's useful

digital_land/utils/add_endpoints_utils.py
```diff
@@ -0,0 +1,708 @@
...
+                    tmp_dir=tmp_dir.absolute(),
+                    org_csv_path=(organisation_dir / "organisation.csv").absolute(),
+                    entity_num_gen=entity_num_gen,
+                )
+
+                all_lookup_entries += resource_lookups
+
+        ctx.obj["NEW_LOOKUPS"] = all_lookup_entries
+
+
```
eveleighoj, line 265: I'm not sure passing this volume of objects through ctx is a good idea. Makes the code hard to read and your never too sure what should be in the ctx

digital_land/utils/add_endpoints_utils.py
```diff
@@ -0,0 +1,708 @@
...
+    new_lookups = ctx.obj["NEW_LOOKUPS"]
+    lookup_csv_path = ctx.obj["PIPELINE_DIR"] / "lookup.csv"
+
+    # create new backups
+    if lookup_csv_path.is_file():
+        with open(lookup_csv_path, "a") as file:
+            for line in new_lookups:
+                file.write(f"{line[0]}\n")
+
+
```
eveleighoj, line 276: a function straight to another function doesn't really make sense

digital_land/utils/add_endpoints_utils.py
```diff
@@ -0,0 +1,708 @@
...
+        entity_num_gen = EntityNumGen(entity_num_state=entity_num_state)
+
+        print("")
+        print("======================================================================")
+        print("New Lookups")
+        print("======================================================================")
+
+        # generate the lookup entries for each new resource
+        dataset_resource_map = collection.dataset_resource_map()
```
eveleighoj, line 244: seems like something that should come from the collection class

digital_land/utils/add_endpoints_utils.py
```diff
@@ -0,0 +1,708 @@
...
+    specification = ctx.obj["SPECIFICATION"]
+    specification_dir = ctx.obj["SPECIFICATION_DIR"]
+    tmp_dir = ctx.obj["TMP_DIR"]
+
+    # get organisation.csv
+    copy_successful = copy_latest_organisation_files_to(organisation_dir)
+
+    if copy_successful:
+        # create and initialise entity number generator
+        entity_num_state = get_entity_number_state(
```
eveleighoj, line 230: this function seems like it could be simpler

digital_land/utils/add_endpoints_utils.py
```diff
@@ -0,0 +1,708 @@
...
+
+    collect(
+        endpoint_path,
+        collection_dir_str,
+        pipeline,
+    )
+
+
+def task_populate_resource_and_log_csvs(ctx):
+    collection = ctx.obj["COLLECTION"]
```
eveleighoj, line 197: no point having both of these

digital_land/utils/add_endpoints_utils.py
```diff
@@ -0,0 +1,708 @@
...
+        csv_file_path,
+        header=0,
+        index_col=False,
+        usecols=cols,
+    )
+
+    collection = Collection()
+    ctx.obj["COLLECTION"] = collection
+
+    for idx, row in csv_file_df.iterrows():
```
eveleighoj, line 113: should it fail if any of the endpoints fail?

digital_land/commands.py
```diff
@@ -403,6 +403,33 @@ def collection_add_source(entry, collection, endpoint_url, collection_dir):
...
+    from digital_land.utils.add_endpoints_utils import (
+        task_preprocess,
+        task_create_source_and_endpoint_entries,
+        task_collect_resources,
+        task_populate_resource_and_log_csvs,
+        task_generate_lookup_entries,
+        task_update_lookup_csv,
+        task_postprocess,
+    )
+
```
eveleighoj, line 421: This is very hard to read as each function is only useful for this purpose. Creating and using classes will be readable as people can read what your doing.

digital_land/utils/add_endpoints_utils.py
eveleighoj: I think we need to add in more standard validation for endpoints and sources. perhaps we can leverage the CsvStore class to create reusable functions for this. It can be a static method attached to one of the classes.

The aim of this is to make it available while using the class (we can include the validations more frequently) This will improve the code base by ensuring that the same methods are used to validate the values.

PR #127, opened 2023-08-03
Fix geojson geometry

kena-SL:
Camden has provided geojson data in the geometry column in a CSV.
(e.g. {"type":"GeometryCollection","geometries":[{"type":"MultiPolygon","coordinates":[[[[-0.185209,51.54305],[-0.185199,51.543051],[-0.18518,51.543054],[-0.185155,51.543057],[-0.185128,51.543062],[-0.185098,51.543066],[-0.185064,51.543071]]]]})

Shapely.wkt.loads() cannot process data in this type and throws exception

This PR handles the exception by converting the code to required format i.e a valid WKT which shapely can process.
An issue with a message "invalid type geojson" is logged
image

Test added

### DIFF
--- a/tests/unit/test_wkt.py
+++ b/tests/unit/test_wkt.py
@@ -167,23 +167,6 @@ def test_wkt_point_missing_values():
     assert issue_type(issues) == "invalid WKT"
 
 
-def test_wkt_geojson():
-    wkt = WktDataType()
-    issues = IssueLog()
-
-    value = (
-        '{"type":"MultiPolygon","coordinates":[[[[-0.143512,51.558395], '
-        "[-0.143592,51.558364],[-0.143665,51.558334],[-0.143677,51.558329], "
-        "[-0.143698,51.558348],[-0.143702,51.558352],[-0.143686,51.558392],"
-        "[-0.143673,51.558398],[-0.143658,51.558404],[-0.143558,51.558442],"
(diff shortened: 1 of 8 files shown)


tests/unit/test_wkt.py
```diff
@@ -167,6 +167,23 @@ def test_wkt_point_missing_values():
     assert issue_type(issues) == "invalid WKT"
 
 
+def test_wkt_geojson():
```
mdsmith-scottlogic, line 170: Its nice to see a test here, I'm not immediately seeing the connection between the change to `parse_wkt` and the test of `wkt.normalize` though. Is it legit to just test the parse function directly?

kena-SL: parse_wkt() function returns Multipolygon object which I am not sure how can be compared with expected value.

normalise() function uses parse_wkt() and processes the Multipolygon object returned by it(i.e. sets precision, checks if its valid etc). 
It returns a string type to which I have compared the expected value.

mdsmith-scottlogic: Ok. 

Maybe testing for the change in sematics from throwing an exception to returning `(geometry, "invalid type geojson")` might have been sufficient but LGTM :-)

mdsmith-scottlogic (approved):

mdsmith-scottlogic (approved):

eveleighoj (approved):

PR #128, opened 2023-08-10
Process type geometrycollection

kena-SL:
Card linked: https://trello.com/c/sKj1EK3Y/303-update-pipeline-to-handle-geometry-type-geometrycollection

This PR fixes the issue of GeometryCollection type throwing unexpected type error
It updated the datatype/wkt file to work with geometry of type GeometryCollection

Condition to check for GeometryCollection type added in:
1. parse_wkt() function which extracts the first_point.
2. make_multipolygon() function which uses unary_union to to compute the union of a various type of geometries and return multipolygon object.

Test verifies that output is as expected and there are no issues reported("Unexpected geom type" issue was being raised before the fix).


GeorgeGoodall (approved):
looks good :)

PR #129, opened 2023-08-15
Empty geometry column fix

kena-SL:
Card linked: https://trello.com/c/xA5vTiKf/312-handle-empty-geometry-column

Southwark has added an empty geometry column in their data.
While processing it in the phase/map file, the data looks something like ('WKT': 'Multipolygon()','geometry':'')
Currently when assigning data first the geometry column is loaded with wkt data and then it is wiped off when the empty geometry column is processed.

This PR modifies the assignment operation and check for data being empty before it is assigned. Hence it retains the geometry loaded by the wkt column.

The test adds an empty geometry column and checks if the fix has retained the geometry data.

### DIFF
--- a/digital_land/phase/map.py
+++ b/digital_land/phase/map.py
@@ -73,7 +73,10 @@ class MapPhase(Phase):
                 if headers[header] == "IGNORE":
                     continue
 
-                o[headers[header]] = row.get(header, "")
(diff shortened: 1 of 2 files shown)


ssadhu-sl (approved):
looks good!

PR #133, opened 2023-09-08
Raising missing organisation issue during pipeline processing

ssadhu-sl:
Reference ticket: https://trello.com/c/M1W1kelh/447-raising-an-issue-for-missing-organisation-at-organisation-phase-itself

During the processing of an endpoint, we have no means to know if we encounter a Missing Organisation issue. Adding this to the issues list raised.

### DIFF
--- a/digital_land/phase/organisation.py
+++ b/digital_land/phase/organisation.py
@@ -18,9 +18,10 @@ class OrganisationPhase(Phase):
                 self.issues.line_number = block["line-number"]
                 self.issues.entry_number = block["entry-number"]
 
+            organisation_value = row["organisation"]
             row["organisation"] = self.organisation.lookup(row["organisation"])
             if not row["organisation"]:
                 self.issues.log_issue(
-                    "organisation", "missing organisation", row["organisation"]
+                    "organisation", "invalid organisation", organisation_value
                 )
             yield block
--- a/digital_land/phase/organisation.py
+++ b/digital_land/phase/organisation.py
@@ -6,7 +6,7 @@ class OrganisationPhase(Phase):
     lookup the organisation
     """
 
-    def __init__(self, issues=None, organisation={}):
+    def __init__(self, organisation={}, issues=None):
         self.organisation = organisation
         self.issues = issues
 
--- a/digital_land/phase/organisation.py
+++ b/digital_land/phase/organisation.py
@@ -6,11 +6,21 @@ class OrganisationPhase(Phase):
     lookup the organisation
     """
 
-    def __init__(self, organisation={}):
+    def __init__(self, issues=None, organisation={}):
         self.organisation = organisation
+        self.issues = issues
 
     def process(self, stream):
         for block in stream:
             row = block["row"]
+            if self.issues:
+                self.issues.resource = block["resource"]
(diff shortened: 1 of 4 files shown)


digital_land/phase/organisation.py
```diff
@@ -6,11 +6,21 @@ class OrganisationPhase(Phase):
     lookup the organisation
     """
 
-    def __init__(self, organisation={}):
+    def __init__(self, issues=None, organisation={}):
```
eveleighoj, line 9: Add this at the end after organisation, it will stop any errors from mixed up order happening

digital_land/phase/organisation.py
```diff
@@ -6,11 +6,21 @@ class OrganisationPhase(Phase):
...
             row = block["row"]
+            if self.issues:
+                self.issues.resource = block["resource"]
+                self.issues.line_number = block["line-number"]
+                self.issues.entry_number = block["entry-number"]
+
             row["organisation"] = self.organisation.lookup(row["organisation"])
+            if not row["organisation"]:
+                self.issues.log_issue(
+                    "organisation", "missing organisation", row["organisation"]
```
eveleighoj, line 24: Assuming this is descriptive enough for swati then it should be fine!

eveleighoj (changes requested):

tests/unit/test_organisation.py
eveleighoj, line 1: We went through this in the meeting but thought I'd put it in a comment:
- move tests to tests/unit/phase/organisation.py as your testing the organisation phase not the organisation class
- as this. is a unit test we should mock the response from the Organisation class, your code doesn't impact what that response will be only reacts based on what's returned

ssadhu-sl: Changes done as discussed, the data from Organisation class is coming from a csv and hence I have kept it like that. The data can be obtained from the csv file we have mocked in the test folder.

tests/unit/phase/test_organisation.py
```diff
@@ -0,0 +1,40 @@
+from digital_land.organisation import Organisation
+from digital_land.phase.organisation import OrganisationPhase
+from digital_land.log import IssueLog
+
+
+def test_process_with_valid_organisation():
+    organisation = Organisation(
```
eveleighoj, line 7: this should be mocked rather than using actual data. Especially for a unit test as in theory they assume that there's no access to a file system. Consider something like 

    def test_process_with_valid_organisation(mocker):

        mocker.patch(
            "digital_land.phase.organisation.Organisation.organisation.lookup",
            return_value='valid_org'
        )
    
This means that the test doesn't have to load data in from the file system and doesn't rely on dependencies outside of your code.

If you wanted to write a test including the organisation class it should be an integration test and I would advice creating your own organisation.csv in temporary location using tmp_path

eveleighoj (changes requested):

tests/unit/phase/test_organisation.py
ssadhu-sl: I was under the impression we test the entire workflow of organisation, but you are right, this is a unit test so should mock the main organisation.
Done the changes now

eveleighoj (approved):

PR #134, opened 2023-09-13
Feat/add endpoints and lookups cmd 2

morgan-sl:
Additional validation / fixes required to resolve issues found during testing of BFL automated endpoint addition.

### DIFF
--- a/digital_land/pipeline.py
+++ b/digital_land/pipeline.py
@@ -328,6 +328,7 @@ class Lookups:
         self.entity_num_gen = EntityNumGen()
 
     def add_entry(self, entry):
+        self.validate_entry(entry)
         self.entries.append(entry)
 
     def load_csv(self, lookups_path=None):
@@ -373,12 +374,22 @@ class Lookups:
                     entry["entity"] = self.entity_num_gen.next()
                 writer.writerow(entry)
 
-    @staticmethod
-    def validate_entry(entry):
+    # @staticmethod
+    def validate_entry(self, entry):
         expected_fields = ["prefix", "organisation", "reference"]
         for field in expected_fields:
             # ensures minimum expected fields exist and are not empty strings
             if not entry.get(field, ""):
                 raise ValueError()
+
+        if len(self.entries) > 0:
+            existing_prefixes = len(
+                [1 for item in self.entries if item["prefix"] == entry["prefix"]]
+            )
+
+            if existing_prefixes == 0:
+                raise ValueError(
+                    f"ERROR: {entry['prefix']} is not expected dataset for this pipeline"
+                )
 
     # I'm not sure we need this class or if we do it should be an iterator then can be used to iterate through entity numbers by dataset
--- a/digital_land/pipeline.py
+++ b/digital_land/pipeline.py
@@ -380,7 +380,7 @@ class Lookups:
         for field in expected_fields:
             # ensures minimum expected fields exist and are not empty strings
             if not entry.get(field, ""):
-                raise ValueError()
+                raise ValueError(f"ERROR: expected {field} not found in lookup entry")
 
         if len(self.entries) > 0:
             existing_prefixes = len(
--- a/digital_land/pipeline.py
+++ b/digital_land/pipeline.py
@@ -347,7 +347,7 @@ class Lookups:
                 [
                     int(entry["entity"])
                     for entry in self.entries
-                    if entry["prefix"] == prefix
+                    if (entry["prefix"] == prefix) and (entry.get("entity", None))
                 ]
             )
         return ret_val
--- a/digital_land/pipeline.py
+++ b/digital_land/pipeline.py
@@ -327,8 +327,17 @@ class Lookups:
         self.schema = Schema("lookup")
         self.entity_num_gen = EntityNumGen()
 
-    def add_entry(self, entry):
-        self.validate_entry(entry)
+    def add_entry(self, entry, is_new_entry=True):
+        """
+        is_new_entry is an addition to allow for backward compatibility.
+        Older lookups may not be valid in accordance with the current
+        minimal column requirements
+        :param entry:
+        :param is_new_entry:
+        :return:
+        """
+        if is_new_entry:
+            self.validate_entry(entry)
         self.entries.append(entry)
 
     def load_csv(self, lookups_path=None):
@@ -338,7 +347,7 @@ class Lookups:
         lookups_path = lookups_path or self.lookups_path
         reader = csv.DictReader(open(lookups_path, newline=""))
         for row in reader:
-            self.add_entry(row)
+            self.add_entry(row, is_new_entry=False)
 
     def get_max_entity(self, prefix):
         if len(self.entries) == 0:
--- a/tests/unit/test_lookups.py
+++ b/tests/unit/test_lookups.py
@@ -1,0 +1,230 @@
+import pytest
+import os
+import csv
+
+from digital_land.pipeline import Lookups
+
+
+@pytest.fixture
+def pipeline_dir(tmp_path):
+    pipeline_dir = os.path.join(tmp_path, "pipeline")
+    os.makedirs(pipeline_dir, exist_ok=True)
+
+    # create lookups
+    row = {
+        "prefix": "ancient-woodland",
+        "resource": "",
+        "organisation": "local-authority-eng:ABC",
+        "reference": "ABC_0001",
+        "entity": "1234567",
+    }
+    fieldnames = row.keys()
+
+    with open(os.path.join(pipeline_dir, "lookup.csv"), "w") as f:
+        dictwriter = csv.DictWriter(f, fieldnames=fieldnames)
+        dictwriter.writeheader()
+        dictwriter.writerow(row)
+
+    return pipeline_dir
+
+
+@pytest.fixture
+def empty_pipeline_dir(tmp_path):
+    pipeline_dir = os.path.join(tmp_path, "pipeline")
+    os.makedirs(pipeline_dir, exist_ok=True)
+
+    return pipeline_dir
+
+
+def test_lookups_load_csv_success(
+    pipeline_dir,
+):
+    """
+    test csv loading functionality
+    :param pipeline_dir:
+    :return:
+    """
+    lookups = Lookups(pipeline_dir)
+    lookups.load_csv()
+
+    expected_num_entries = 1
+    assert len(lookups.entries) == expected_num_entries
+
+
+def test_lookups_load_csv_failure(
+    empty_pipeline_dir,
+):
+    """
+    test csv loading functionality when file absent
+    :param pipeline_dir:
+    :return:
+    """
+    lookups = Lookups(empty_pipeline_dir)
+
+    with pytest.raises(FileNotFoundError) as fnf_err:
+        lookups.load_csv()
+
+    expected_exception = "No such file or directory"
+    assert expected_exception in str(fnf_err.value)
+
+
+def test_lookups_get_max_entity_success(
+    pipeline_dir,
+):
+    """
+    test entity num generation functionality
+    :param pipeline_dir:
+    :return:
+    """
+    pipeline_name = "ancient-woodland"
+
+    lookups = Lookups(pipeline_dir)
+    lookups.load_csv()
+
+    expected_next_entity_num = 1
+    assert lookups.entity_num_gen.next() is expected_next_entity_num
+
+    max_entity_num = lookups.get_max_entity(pipeline_name)
+    lookups.entity_num_gen.state["current"] = max_entity_num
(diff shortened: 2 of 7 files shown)


tests/unit/test_lookups.py
```diff
@@ -0,0 +1,230 @@
+import pytest
```
eveleighoj, line 1: unit and integration tests should replicate code structure these shouldn't be in a random file and should match where the lookups class is. i.e. it should be in 

`tests/unit/pipeline.py`

tests/unit/test_lookups.py
```diff
@@ -0,0 +1,230 @@
...
+import os
+import csv
+
+from digital_land.pipeline import Lookups
+
+
+@pytest.fixture
```
eveleighoj, line 8: unit tests should not be reliant on a file system as a dependency. Either it should be an integration test or by mocking.

tests/unit/test_lookups.py
```diff
@@ -0,0 +1,230 @@
...
+def empty_pipeline_dir(tmp_path):
+    pipeline_dir = os.path.join(tmp_path, "pipeline")
+    os.makedirs(pipeline_dir, exist_ok=True)
+
+    return pipeline_dir
+
+
```
eveleighoj, line 38: Move load_csv testing to integration tests as it requires a file system

tests/unit/test_lookups.py
```diff
@@ -0,0 +1,230 @@
...
+    """
+    lookups = Lookups(empty_pipeline_dir)
+
+    with pytest.raises(FileNotFoundError) as fnf_err:
+        lookups.load_csv()
+
+    expected_exception = "No such file or directory"
+    assert expected_exception in str(fnf_err.value)
+
+
```
eveleighoj, line 70: should be able to run a test that doesn't need a file system by just adding them to the class

tests/unit/test_lookups.py
```diff
@@ -0,0 +1,230 @@
...
+    assert lookups.entity_num_gen.next() is expected_next_entity_num
+
+    max_entity_num = lookups.get_max_entity(pipeline_name)
+    lookups.entity_num_gen.state["current"] = max_entity_num
+    lookups.entity_num_gen.state["range_max"] = max_entity_num + 10
+
+    expected_next_entity_num = 1234568
+    assert lookups.entity_num_gen.next() == expected_next_entity_num
+
+
```
eveleighoj, line 94: not sure you need the pipeline_dir for this

tests/unit/test_lookups.py
```diff
@@ -0,0 +1,230 @@
...
+    pipeline_dir,
+):
+    """
+    test csv validate_entry functionality for various errors
+    :param pipeline_dir:
+    :return:
+    """
+    lookups = Lookups(pipeline_dir)
+    lookups.load_csv()
+
```
eveleighoj, line 127: I would advise parameterising this test. instead of repeating for each entry

tests/unit/test_lookups.py
```diff
@@ -0,0 +1,230 @@
...
+        "prefix": "",
+        "resource": "",
+        "organisation": "",
+        "reference": "",
+        "entity": "",
+    }
+    with pytest.raises(ValueError):
+        lookups.validate_entry(entry)
+
+
```
eveleighoj, line 163: doesn't need pipeline_dir

tests/unit/test_lookups.py
```diff
@@ -0,0 +1,230 @@
...
+        "resource": "",
+        "organisation": "government-organisation:D1342",
+        "reference": "1",
+        "entity": "",
+    }
+
+    lookups.add_entry(entry)
+
+
+def test_lookups_add_entry_failure(
```
eveleighoj, line 186: Doesn't need pipeline_dir

tests/unit/test_lookups.py
```diff
@@ -0,0 +1,230 @@
...
+    pipeline_dir,
+):
+    """
+    test csv add_entry functionality for validation errors
+    :param pipeline_dir:
+    :return:
+    """
+    lookups = Lookups(pipeline_dir)
+    lookups.load_csv()
+
```
eveleighoj, line 196: again test can be paramterised

eveleighoj (changes requested):

tests/unit/test_pipeline.py
```diff
@@ -93,3 +95,150 @@ def test_migrate():
...
+    lookups.entity_num_gen.state["current"] = max_entity_num
+    lookups.entity_num_gen.state["range_max"] = max_entity_num + 10
+    expected_entity_num = 12345
+
+    assert lookups.entity_num_gen.next() == expected_entity_num
+
+
+def test_lookups_validate_entry_success():
+    """
+    test csv validate_entry functionality
```
eveleighoj, line 134: In all these descriptions it mentions a csv I think this is misleading so may be worth removing the word csv.

tests/unit/test_pipeline.py
```diff
@@ -93,3 +95,150 @@ def test_migrate():
...
+    lookups.validate_entry(entry)
+
+
+def test_lookups_validate_entry_failure():
+    """
+    test csv validate_entry functionality for various errors
+    :return:
+    """
+    lookups = Lookups("")
+    invalid_entries = [
```
eveleighoj, line 156: this is better as less duplication you can use 
`@pytest,mark.parametise`
to remove them from inside the test and the test willl run three separate times. This can help when understanding what cause a test to fail. see [here](https://docs.pytest.org/en/7.3.x/how-to/parametrize.html) for some documentation

tests/unit/test_pipeline.py
```diff
@@ -93,3 +95,150 @@ def test_migrate():
...
+
+    entry = {
+        "prefix": "ancient-woodland",
+        "resource": "",
+        "organisation": "government-organisation:D1342",
+        "reference": "1",
+        "entity": "",
+    }
+
+    lookups.validate_entry(entry)
```
eveleighoj, line 147: There should be an assert statement here, making sure that the response is True

`assert lookups.validate_entry(entry)`

tests/unit/test_map.py
```diff
@@ -75,3 +75,46 @@ def test_map_empty_geometry_column():
         "entry-date,geometry,legislation,name,notes,organisation,point,prefix,"
         "reference,start-date\r\n,,,,,,MULTIPOLYGON(),,,,,,,,\r\n"
     )
+
+
```
eveleighoj, line 79: You shouldn't be working on multiple bits of work in one PR. the mapping PR shouldn't really be touching any of the same code. Also this is testing the map phase so should be in tests/unit/phase/map.py

tests/integration/test_pipeline.py
```diff
@@ -320,3 +356,125 @@ def test_concatenations_when_csv_contains_endpoints_and_resources(tmp_path):
...
+    assert lookups.entity_num_gen.next() is expected_next_entity_num
+
+    max_entity_num = lookups.get_max_entity(pipeline_name)
+    lookups.entity_num_gen.state["current"] = max_entity_num
+    lookups.entity_num_gen.state["range_max"] = max_entity_num + 10
+
+    expected_next_entity_num = 1234568
+    assert lookups.entity_num_gen.next() == expected_next_entity_num
+
+
```
eveleighoj, line 416: adding an entry doesn't require a csv and hence can be a unit test. this feels like duplication

tests/integration/test_pipeline.py
```diff
@@ -320,3 +356,125 @@ def test_concatenations_when_csv_contains_endpoints_and_resources(tmp_path):
...
+    """
+    lookups = Lookups(empty_pipeline_dir)
+
+    with pytest.raises(FileNotFoundError) as fnf_err:
+        lookups.load_csv()
+
+    expected_exception = "No such file or directory"
+    assert expected_exception in str(fnf_err.value)
+
+
```
eveleighoj, line 392: duplication of functionality in unit test

digital_land/pipeline.py
```diff
@@ -327,7 +327,19 @@ def __init__(self, directory=None) -> None:
         self.schema = Schema("lookup")
         self.entity_num_gen = EntityNumGen()
 
-    def add_entry(self, entry):
+    def add_entry(self, entry, is_new_entry=True):
```
eveleighoj, line 330: should 
`is_new_entry` 
actually be 
`validate`
as it's actually turning validation on and off

digital_land/phase/map.py
```diff
@@ -6,8 +6,13 @@
 normalise_pattern = re.compile(r"[^a-z0-9-_]")
 
 
+def replace_underscores(name: str):
```
eveleighoj, line 9: This looks right but again shouldn't be in this PR. also I'm not sure we need another function to remove underscores when it's one line it could just be in the normalise function with an extra test

eveleighoj (changes requested):

tests/integration/test_pipeline.py
morgan-sl: It is very similar to the unit test.
The idea was that this one tests the functionality using file I/O.

tests/integration/test_pipeline.py
morgan-sl: This test isn't in the unit test anymore.
Is this an old comment?

digital_land/pipeline.py
morgan-sl: This is just a matter of opinion.
One is a property of the entry, the other is an instruction to the method.
I'd prefer attributing the parameter rather than instructing the method because I'd rather not declare that validation is optional. It's really just a quirk for backward compatibility.
However, if changing it gets it 'passed' then I'll just change it.

tests/integration/test_pipeline.py
eveleighoj: see above comment

tests/integration/test_pipeline.py
eveleighoj: It appears above the line I commented on for whatever reason. I was talking about 
`test_lookups_get_max_entity_success`
My comment below was for 
`test_lookups_add_entry_success`
which I believe still has a unit test. My concern is that they're not testing anything here in addition  to what is done by load csv or in the unit tests. Unit tests are what we should aim for.

digital_land/pipeline.py
eveleighoj: Happy to leave as is my concern is that it is more difficult to understand. but I do agree it's an opinion!

tests/unit/test_pipeline.py
eveleighoj: Any thoughts on this?

morgan-sl: Well I've accepted past comments that I've used too many fixtures and it made the code hard to read, coz you have to go up and down the file.
I guess I figure this has everything in one place. After all, parameterizing (usually) is just a looping mechanism controlled by the Test library.
Typically, I would always go for least complexity.

tests/integration/test_pipeline.py
morgan-sl: I agree they seem very similar.
I put both tests in - one for in-memory, one to ensure it's getting the last entity number from teh correct place in a file.
That is the only difference, though.

tests/integration/test_pipeline.py
morgan-sl: Hehe.
This bit's got me confused.
I'll wait to see if I've addressed the correct 'above' concern.

tests/unit/test_pipeline.py
eveleighoj: I don't think it makes a huge difference. it would be a simple decorator function and the console output we be improved as it runs each as an individual test. running the below if would difficult to tell which of the invalid entries was causing the test to fail and a dev would likely need to figure out which one it is before debugging that one. 

It's not a massive deal but as a python dev who uses pytest I would expect this to be a prime example of parameterisation is useful.

tests/integration/test_pipeline.py
eveleighoj: I think remove them. the load_csv tests will test reading in the data. The unit test will test getting the max entity out.

morgan-sl: done

tests/unit/test_pipeline.py
eveleighoj: Looks good! it can also be applied to 

`test_lookups_validate_entry_failure`

tests/unit/test_map.py
```diff
@@ -37,41 +37,3 @@ def test_map_headers_column_clash():
     m = MapPhase(["One"], {"une": "One", "ein": "One"})
     output = TestPipeline(m, "une,ein\r\n1,2\r\n")
     assert output == "One\r\n1\r\n"
-
```
eveleighoj, line 40: looks like you've deleted a test that was already in their

eveleighoj (approved):

PR #136, opened 2023-09-15
Replace underscores in column names with hyphens

morgan-sl:
made changes to the MapPhase to perform column name transform (underscore to hyphen) without the need for manual intervention

### DIFF
--- a/digital_land/phase/map.py
+++ b/digital_land/phase/map.py
@@ -6,8 +6,13 @@ from .phase import Phase
 normalise_pattern = re.compile(r"[^a-z0-9-_]")
 
 
+def replace_underscores(name: str):
+    return name.replace("_", "-")
+
+
 def normalise(name):
-    return re.sub(normalise_pattern, "", name.lower())
+    new_name = replace_underscores(name)
+    return re.sub(normalise_pattern, "", new_name.lower())
 
 
 class MapPhase(Phase):
@@ -18,7 +23,9 @@ class MapPhase(Phase):
 
     def __init__(self, fieldnames, columns={}, log=None):
         self.columns = columns
-        self.normalised_fieldnames = {normalise(f): f for f in fieldnames}
+        self.normalised_fieldnames = {
+            normalise(f): replace_underscores(f) for f in fieldnames
+        }
         if not log:
             log = ColumnFieldLog()
         self.log = log
--- a/digital_land/phase/map.py
+++ b/digital_land/phase/map.py
@@ -6,12 +6,8 @@ from .phase import Phase
 normalise_pattern = re.compile(r"[^a-z0-9-_]")
 
 
-def replace_underscores(name: str):
-    return name.replace("_", "-")
-
-
 def normalise(name):
-    new_name = replace_underscores(name)
+    new_name = name.replace("_", "-")
     return re.sub(normalise_pattern, "", new_name.lower())
 
 
@@ -23,9 +19,7 @@ class MapPhase(Phase):
 
     def __init__(self, fieldnames, columns={}, log=None):
         self.columns = columns
-        self.normalised_fieldnames = {
-            normalise(f): replace_underscores(f) for f in fieldnames
-        }
+        self.normalised_fieldnames = {normalise(f): f for f in fieldnames}
         if not log:
             log = ColumnFieldLog()
         self.log = log
--- a/tests/unit/phase/map.py
+++ b/tests/unit/phase/map.py
@@ -1,10 +1,12 @@
 #!/usr/bin/env -S pytest -svv
+import pytest
 
 from io import StringIO
 from digital_land.pipeline import run_pipeline
 from digital_land.phase.load import LoadPhase
 from digital_land.phase.parse import ParsePhase
 from digital_land.phase.map import MapPhase
+from digital_land.phase.map import normalise as map_normalise
 from digital_land.phase.save import SavePhase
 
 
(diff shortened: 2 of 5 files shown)


digital_land/phase/map.py
```diff
@@ -6,8 +6,13 @@
 normalise_pattern = re.compile(r"[^a-z0-9-_]")
 
 
+def replace_underscores(name: str):
```
eveleighoj, line 9: this probably doesn't need an extra function, the function name is longer than the code it actually runs.

tests/unit/test_map.py
```diff
@@ -75,3 +75,46 @@ def test_map_empty_geometry_column():
         "entry-date,geometry,legislation,name,notes,organisation,point,prefix,"
         "reference,start-date\r\n,,,,,,MULTIPOLYGON(),,,,,,,,\r\n"
     )
+
```
eveleighoj, line 78: this is testing the Map class so it should be in:
`tests/unit/phase/map.py`
also I can;t really to the method/function that this is testing from the name. Also the TestPipeline class may not be needed see [here](https://github.com/digital-land/digital-land-python/blob/main/tests/unit/phase/test_organisation.py) you can just feed in a dict pretty much and it's easier to read

eveleighoj (changes requested):

digital_land/phase/map.py
```diff
@@ -18,7 +23,9 @@ class MapPhase(Phase):
 
     def __init__(self, fieldnames, columns={}, log=None):
         self.columns = columns
-        self.normalised_fieldnames = {normalise(f): f for f in fieldnames}
+        self.normalised_fieldnames = {
+            normalise(f): replace_underscores(f) for f in fieldnames
```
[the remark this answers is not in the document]
morgan-sl: It does look weird, but it's building a dictionary where both the keys and values are fieldnames.
From what your suggesting above it would become 
normalise(f): normalise(f) for f...
I'll amend accordingly.

digital_land/phase/map.py
morgan-sl: I was being cautious as I'd seen normalise being used for things like urls.
I now appreciate that many modules have a normalize method, so the impact of changing map's normalise is probably localized to our need - affecting only column-names.
I'll extend normalise() as advised.

morgan-sl: I found that changing it to one function broke a lot of existing tests.
By leaving it as 2 functions, it works as we need and existing functionality needs.
I know it looks weird, but I feel that this is the best approach.

tests/unit/test_map.py
morgan-sl: I think I've already commented on this, but there is already a unit/test_map class.
I put the code in there to be consistent with what has been done previously.
Would it not be confusing to have Map class tests in 2 locations.
If we move this one, then we have to move many others, too.

morgan-sl: I've changed the function names.
hope they're a bit more understandable.

digital_land/phase/map.py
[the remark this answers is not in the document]
morgan-sl: Due to my comment above, I'll leave this the way it is for now.

digital_land/phase/map.py
eveleighoj: the only tests that are failing are the ones you've written. I think the issue with them is that the mapping phase itself doesn't remove columns so the. old columns are left in

morgan-sl:
I think the key statement in your review was:

_So to be clear this should be:_
_normalise(f): f for f in fieldnames_
**_Otherwise you altering the actual fieldname that it's mapped to which we don't want to do. What we're doing is_ _making sure that it doesn't matter in the matching phase._**

I think I assumed mapping was to transform columns to our standard.
I've made the changes you suggested (all the tests passing) but, if I'm honest, I don't really get what the test is telling me.
Does the output look like what you'd expect?

If so, I understand what you mean by
_I think the issue with them is that the mapping phase itself doesn't remove columns so the. old columns are left in_

tests/unit/phase/map.py
```diff
@@ -75,3 +77,61 @@ def test_map_empty_geometry_column():
         "entry-date,geometry,legislation,name,notes,organisation,point,prefix,"
         "reference,start-date\r\n,,,,,,MULTIPOLYGON(),,,,,,,,\r\n"
     )
+
+
```
eveleighoj, line 81: these tests still need moving to the appropriate location in the test directory. Just because the file exists does not mean that we shouldn't focus on improving the tests directory structure.

tests/unit/phase/map.py
```diff
@@ -1,10 +1,12 @@
 #!/usr/bin/env -S pytest -svv
+import pytest
 
 from io import StringIO
 from digital_land.pipeline import run_pipeline
 from digital_land.phase.load import LoadPhase
 from digital_land.phase.parse import ParsePhase
 from digital_land.phase.map import MapPhase
+from digital_land.phase.map import normalise as map_normalise
```
eveleighoj, line 9: why has this changed name? It doesn't seem necessary.

tests/unit/phase/map.py
morgan-sl: I thought it would be a separate ticket - so as not to bind errors related to moving, to this specific task.

tests/unit/phase/map.py
morgan-sl: I'll change it.

eveleighoj (approved):

PR #139, opened 2023-09-22
Refines invalid geometry issue 

kena-SL:
Displays issue "geometry is too precise - fixed" when the geometry provided has precision over 6 decimal points.


hyperoceanic:
@kena-SL - please hit me up next week so we can update the docs on this.

digital_land/datatype/wkt.py
```diff
@@ -209,24 +214,27 @@ def normalise(self, value, default="", issues=None):
...
+                issues.log("geometry is too precise - fixed", "")
+
             geometry = shapely.wkt.loads(_wkt)
             validity = False
             i = 0
             while i < 3 and not validity:
                 geometry, issue = normalise_geometry(geometry)
                 if issue:
-                    issues.log("invalid geometry", issue)
+                    issues.log("invalid geometry provided - fixed", issue)
```
eveleighoj, line 227: This is where stuff gets a bit wild. we may need to rethink how we record these issues as it's a loop. specifically in this case we are primarily fixing errors introduced by attempting to reduce the precision, at this stage we have fix the issue caused by reducing precision but we will dump it out again below. I'm not against describing that we fixed this but I'm worried it could be misleading when examining the issue log

digital_land/datatype/wkt.py
```diff
@@ -209,24 +214,27 @@ def normalise(self, value, default="", issues=None):
...
+                _wkt, issue = dump_wkt(geometry)
                 geometry = shapely.wkt.loads(_wkt)
                 validity = geometry.is_valid
                 i += 1
 
         if not geometry:
             return default
 
-        return dump_wkt(geometry)
+        return dump_wkt(geometry)[0]
```
eveleighoj, line 240: we could update this area. Right now as you can see if the geometry is still invalid we still actually include it. I would suggest a simple .is_valid check to see if it is finally valid and then to remove it if it's not and record an issue.

digital_land/datatype/wkt.py
```diff
@@ -182,10 +183,14 @@ def normalise_geometry(geometry, simplification=0.000005):
...
 def dump_wkt(geometry, precision=6, dimensions=2):
+    try:
+        current_precision = len(str(x).split(".")[1]) if "." in str(x) else 0
+    except Exception:
+        current_precision = 0
     wkt = shapely.wkt.dumps(
         geometry, rounding_precision=precision, output_dimension=dimensions
     )
-    return wkt.replace(", ", ",")
+    return wkt.replace(", ", ","), True if current_precision > 6 else False
```
eveleighoj, line 193: I'm worried that this is going to be very late in the process, i.e. we've made a lot of changes by now so the precision issue could be caused by our transformations etc. rather than them providing a geometry that's too precise this issue would also show that we transformed it to WGS84 and it's too precise, which while it is useful still to us it may confuse them if we showed them the issue. 

We probably need a check which happens at the start to inform us whether they are providing the geometry as too precise, and maybe one nearer the end that tells us if precision introduced during processing? Not 100% sure if we need two or not.

I'm also wondering whether we need this to happen in the dump_wkt function if it's something that can be checked somewhere else in the wkt.normalise method

digital_land/phase/combine.py
```diff
@@ -14,7 +14,7 @@ def combine_geometries(wkts, precision=6):
     union = unary_union(geometries)
     if not isinstance(union, MultiPolygon):
         union = MultiPolygon([union])
-    return dump_wkt(union, precision=precision)
+    return dump_wkt(union, precision=precision)[0]
```
eveleighoj, line 17: another arguement for moving the issue checking outside of this function

eveleighoj (changes requested):

digital_land/datatype/wkt.py
kena-SL: Done

digital_land/datatype/wkt.py
kena-SL: I have added this at the start of wkt.normalise and removed it from dump_wkt as it being called multiple times.
Currently it just says 'geometry provided is too precise - fixed' for when geometry is greater than 6 d.p. 
Can also rename to 'handled by pipeline' instead of 'fixed'.

How can we check at the end to make sure precision has not been introduced?

PR #140, opened 2023-09-27
fix/shapely_deprecation_warning

morgan-sl:
Amended code to use the modern approach to accessing parts of multipart geometry definitions using the Shapely library.


CarlosCoelhoSL (approved):
lgtm

PR #145, opened 2023-10-06
Import local authority geometry into pipeline for boundary check

CarlosCoelhoSL:
This PR adds functionality to process and load LA geometry data into the harmonise phase. This geometry data will be download during the make init process, using changes found in this PR: pr here.

The downloaded LA geometry data is pulled directly from source (Office for National Statistics) rather than our data, as we only have clipped to coastline geometry whereas extent of the realm is best used for point in polygon checks. This means the data must be processed into the correct format for the boundary check, which happens before the pipeline is run using a utl file, where it is then saved as a .csv and accessed when needed.

Checking whether an entity lies within its Local Authority boundary only happens for the brownfield land dataset


digital_land/commands.py
```diff
@@ -140,8 +140,16 @@ def pipeline_run(
     default_values = pipeline.default_values(endpoints=endpoints)
     combine_fields = pipeline.combine_fields(endpoints=endpoints)
 
+    lpa_geometry_path = "var/cache/local-authority.csv"
+    local_authority_path = "var/cache/local-authority.csv"
```
CarlosCoelhoSL, line 144: I've hardcoded these for now, it should be pretty easy to add these as inputs to the pipeline run command, but I'm not sure where that run command is actually triggered

GeorgeGoodall (approved):

PR #147, opened 2023-10-12
data_eng/optimize_dataset_create

morgan-sl:
The purpose of the change is to speed up the creation of dataset sqlite3 files.
The analysis summary follows, which I believe justifies the change.

### Performance Metrics of Optimization Changes
There are three datasets in the listed-building collection:

- listed-building
- listed-building-grade
- listed-building-outline

Each dataset has to process it's corresponding issue log file.

Total size of Issue Logs for each dataset

- listed-building 138,476 KB
- listed-building-grade 1 KB
- listed-building-outline 1,080 KB

Listed building clearly has the most work to do when generating the issue-log table.

The listed-building dataset has the largest transform file, containing 3412530 records. This file was 'cut' into different sizes to get the timing set below.

| Run |       # Records Processed | Current Solution (s) | New Solution (s) |     % | 
|---:|--------------------------:|---------------------:|-----------------:|:------:|
|  1  |                      1000 |                23.80 |             9.57 | 40.21 |
|  2  |                     64000 |                30.39 |            13.81 | 45.44 |
|  3  |                    512000 |                87.02 |            54.45 | 62.57 |
|  4  |                   1024000 |               157.59 |           113.61 | 72.09 |
|  5  |                   2048000 |               312.41 |           235.45 | 75.37 |
|  6  |                   3412530 |               518.65 |           384.26 | 74.09 |

Table shows the number of seconds taken to process a given number of listed-building transform file records (+ issue-log records from above).

### Summary
It’s clear the savings made from optimization diminish with increased transform file size.
The savings reduce to about 25% for very large datasets, but can be as much as 60% for smaller datasets.

It’s worth noting that, although the savings from large datasets like listed-building is small, the savings from processing all collections should be significant enough to see an overall benefit from these changes.

### DIFF
--- a/digital_land/cli.py
+++ b/digital_land/cli.py
@@ -108,7 +108,7 @@ def convert_cmd(input_path, output_path):
 
 @cli.command("dataset-create", short_help="create a dataset from processed resources")
 @click.option("--output-path", type=click.Path(), default=None, help="sqlite3 path")
-@click.option("--max-batch-size", type=click.INT)
+@click.option("--max-batch-size", type=click.INT, default=10000)
 @click.argument("input-paths", nargs=-1, type=click.Path(exists=True))
 @organisation_path
 @click.pass_context
--- a/digital_land/cli.py
+++ b/digital_land/cli.py
@@ -108,10 +108,13 @@ def convert_cmd(input_path, output_path):
 
 @cli.command("dataset-create", short_help="create a dataset from processed resources")
 @click.option("--output-path", type=click.Path(), default=None, help="sqlite3 path")
+@click.option("--max-batch-size", type=click.INT)
 @click.argument("input-paths", nargs=-1, type=click.Path(exists=True))
 @organisation_path
 @click.pass_context
-def dataset_create_cmd(ctx, input_paths, output_path, organisation_path):
+def dataset_create_cmd(
+    ctx, input_paths, output_path, max_batch_size, organisation_path
+):
     return dataset_create(
         input_paths=input_paths,
         output_path=output_path,
@@ -119,6 +122,7 @@ def dataset_create_cmd(ctx, input_paths, output_path, organisation_path):
         pipeline=ctx.obj["PIPELINE"],
         dataset=ctx.obj["DATASET"],
         specification=ctx.obj["SPECIFICATION"],
+        max_batch_size=max_batch_size,
     )
 
 
--- a/digital_land/commands.py
+++ b/digital_land/commands.py
@@ -249,7 +249,6 @@ def dataset_create(
         specification_dir=None,  # TBD: package should use this specification object
     )
     package.create()
-    package.connect(optimised=True)
 
     for path in input_paths:
         package.load_transformed(path)
@@ -268,8 +267,6 @@ def dataset_create(
         logging.warning("No directory for this dataset in the provided issue_directory")
 
     package.add_counts()
-
-    package.disconnect()
 
 
 def dataset_dump(input_path, output_path):
--- a/digital_land/commands.py
+++ b/digital_land/commands.py
@@ -236,8 +236,8 @@ def dataset_create(
     pipeline,
     dataset,
     specification,
-    max_batch_size,
     issue_dir="issue",
+    max_batch_size=10000,
 ):
     if not output_path:
         print("missing output path", file=sys.stderr)
--- a/digital_land/commands.py
+++ b/digital_land/commands.py
@@ -249,8 +249,11 @@ def dataset_create(
         specification_dir=None,  # TBD: package should use this specification object
     )
     package.create()
+    package.connect(optimised=True)
+
     for path in input_paths:
         package.load_transformed(path)
+
     package.load_entities()
 
     old_entity_path = os.path.join(pipeline.path, "old-entity.csv")
@@ -265,6 +268,8 @@ def dataset_create(
         logging.warning("No directory for this dataset in the provided issue_directory")
 
     package.add_counts()
+
+    package.disconnect()
 
 
 def dataset_dump(input_path, output_path):
--- a/digital_land/commands.py
+++ b/digital_land/commands.py
@@ -236,6 +236,7 @@ def dataset_create(
     pipeline,
     dataset,
     specification,
+    max_batch_size,
     issue_dir="issue",
 ):
     if not output_path:
@@ -246,6 +247,7 @@ def dataset_create(
         dataset,
         organisation=organisation,
         path=output_path,
+        max_batch_size=max_batch_size,
         specification_dir=None,  # TBD: package should use this specification object
     )
     package.create()
--- a/digital_land/package/dataset.py
+++ b/digital_land/package/dataset.py
@@ -1,8 +1,8 @@
 import csv
 import json
 import logging
-import re
 from decimal import Decimal
+from pathlib import Path
 
 import shapely.wkt
 
@@ -123,11 +123,13 @@ class DatasetPackage(SqlitePackage):
         row = self.entity_row(facts)
         row = self.migrate_entity(row)
         if row:
-            self.insert("entity", self.entity_fields, row)
+            entity_fields = [
+                field for field in self.entity_fields if not field.endswith("-geom")
+            ]
+            self.add_table_values(entity_fields, row)
 
     def load_old_entities(self, path):
         """load the old-entity table"""
-
         fields = self.specification.schema["old-entity"]["fields"]
         entity_min = self.specification.schema[self.dataset].get("entity-minimum")
         entity_max = self.specification.schema[self.dataset].get("entity-maximum")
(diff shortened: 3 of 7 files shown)


digital_land/package/sqlite.py
```diff
@@ -46,18 +49,101 @@ def spatialite(self, path=None):
...
+            condition = " ON CONFLICT DO NOTHING"
+            sql_stmt = f"INSERT OR REPLACE INTO {table}({table_cols}) VALUES({placeholders}){condition};"
+        else:
+            sql_stmt = (
+                f"INSERT OR REPLACE INTO {table}({table_cols}) VALUES({placeholders});"
+            )
+
+        try:
+            # print(f">>> table: {table} - table_values: {sys.getsizeof(self.table_values)}")
+            self.cursor.executemany(sql_stmt, self.table_values)
```
mdsmith-scottlogic, line 133: 👍 not present in the codebase thus far.

morgan-sl: good spot.
thanks

digital_land/package/sqlite.py
```diff
@@ -46,18 +49,100 @@ def spatialite(self, path=None):
...
         self.connection.close()
 
+    def create_cursor(self, optimised=False):
+        if self.cursor:
+            self.cursor.close()
+            self.cursor = None
+
+        self.cursor = self.connection.cursor()
+        if optimised:
+            self.cursor.execute("PRAGMA synchronous = OFF")
```
eveleighoj, line 80: PARGMA synchronous = OFF and PRAGMA journal_mode = OFF were already being used. If this code is being used outside of the command (which is it) those processes will become less efficient as they're no longer set. 

Check the digital-land-builder repo for another example of where the sqlite package is being used.  by applying this change I imagine those processes would slow down.

digital_land/package/sqlite.py
```diff
@@ -46,18 +49,100 @@ def spatialite(self, path=None):
...
+            logging.error("Exception: %s" % (error.__class__))
+            logging.error("Sqlite3 error: %s" % (" ".join(error.args)))
+            logging.info("Sqlite3 cmd: %s" % (sql_stmt))
+            sys.exit(3)
+
+    def add_table_values(self, fields, row):
+        fields = [field for field in fields if not field.endswith("-geom")]
+
+        table_vals = tuple([self.colvalue2(row, field) for field in fields])
+        self.table_values.append(table_vals)
```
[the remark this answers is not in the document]
morgan-sl: I had the same concern.
I ran some memory tests using listed-building, which has a transformed size of 652MB.
The memory size of table values array was only in the 30s or 40s of MB.
That said, I'm definitely open to testing in a non-prod environment to check such things.
It should be possible to periodically commit a batch of say 10000 records, for example.

digital_land/package/sqlite.py
morgan-sl: Ah ok.
The limit of my visibility was the tests.
I looked for other uses in dl-python, but that was it. So apologies for the assumption here.

The intent was that the DDL statements would not use these flags, because it shouldn't be used for these types of statements.
Maybe the answer is to create a dedicated DDL cursor. Then the general create cursor function can have them in as before.

digital_land/package/sqlite.py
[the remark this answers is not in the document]
eveleighoj: Listed buildings has points for geometries though where as flood risk zones has full geometries I can imagine this impacts it. Also with the inspire polygons on the horizon introducing more memory based restrictions seems dangerous. At the minimum we need to implement batching even if it's a large batch for now. I'm wondering if it needs to be saved in memory here or is it can be directly loaded into the transaction.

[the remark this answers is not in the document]
morgan-sl: I'm leaning toward batching, otherwise we lose the benefit of batch transactions.
We'd just be back to where we started, effectively.
During my analysis, this part has the greatest potential to increase performance.

That said, if we're getting into the realms of processing GB file sizes, we might need to at least start thinking about more robust, enterprise-grade solutions.
You know what they say about rolling your own :).

Anyway, if you're not overly opposed to the current solution being batched, then I'll implement that.

Thanks for the heads up re. FRZ.
I was under the impression listed-building was the principal problem ds, and everything else would only be smaller.

digital_land/package/sqlite.py
eveleighoj: Yeah I think that the cursor needs to at least contain what it did before, even if it's an assumption that its optimised. I would also question whether the begin statement should be in the cursor definition. From my perspective the cursor could be used with or without it. It's the individual function for the query which should decide to use it.

morgan-sl: We're agreed on the first bit.
I'll include that.

Not sure about the second bit, to be honest.
If the cursor client was in charge of the query they run, I would agree with you.
However, the code uses predefined queries in functions and if, as is the case here, the predefined queries are to perform in a uniformly optimal way (i.e., batch commits whenever possible, prepared statements, etc.), then we shouldn't leave the inclusion of the transactional marker to chance.
For example., the DatasetPackage cursor client is only concerned with calling the right sequence of functions and passing the appropriate data. It has no knowledge of whether these should be transactional or not.
I feel your suggestion requires the cursor client to have knowledge of the implementation of the function.

morgan-sl:
Revision II attempt to address many of the concerns raised above.

digital_land/package/sqlite.py
```diff
@@ -46,18 +49,100 @@ def spatialite(self, path=None):
...
+    def create_cursor(self, transactional=False):
+        if self.cursor:
+            self.cursor.close()
+            self.cursor = None
+
+        self.cursor = self.connection.cursor()
+        self.cursor.execute("PRAGMA synchronous = OFF")
+        self.cursor.execute("PRAGMA journal_mode = OFF")
+        if transactional:
+            self.cursor.execute("BEGIN")
```
eveleighoj, line 82: I'm not we need to execute BEGIN here. It makes more sense to me to add the into the functions that are doing the loading rather than here. Also for the batching you've implemented above will it only apply to the first batch? Once the commit is ran?

Also doesn't executemany automatically negate the need for this? As it executes all of the statements in one transaction?

Could we not just leave the insert function the same but run a begin before it then not even need to store anything in a python list? Have you tried this?

morgan-sl: If we add BEGIN to the loading functions, then they contain SQL and functional code.
I'm trying to centralize the SQL code to one place, to make it easier to look after in future.
I used a Transactional param to satisfy your previous wish to have other clients of SQLite package choose whether they use Transactions or not.
The BEGIN is applied at connection-level, not cursor level. We just need to make sure the 2 separate connections (one for the DDL statements, and the other for the INSERTS) can independently control whether they use transactions or not.
The commit does not cancel the Transaction, closing the current connection does.

Execute many does execute all in one transaction, and _may_ apply one if it's not there.
Never a bad thing to be explicit in our intent, though - especially as we've created a connection and cursor at this point.

Yes I tried that and many other permutations.
All modifications were made incrementally (see Trello ticket comments). Simply using a BEGIN does not provide the speed increase that this version does.

digital_land/package/dataset.py
```diff
@@ -119,15 +120,23 @@ def entity_row(self, facts):
             row[fact[1]] = fact[2]
         return row
 
-    def insert_entity(self, facts):
+    def insert_entity(self, facts, insert_rows):
```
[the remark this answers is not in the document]
eveleighoj: Looking at the above it does extend the function quite a lot but I'm not sure the insert_entity function is the right way to simplify it

digital_land/commands.py
```diff
@@ -236,6 +236,7 @@ def dataset_create(
     pipeline,
     dataset,
     specification,
+    max_batch_size,
```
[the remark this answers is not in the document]
morgan-sl: agreed to make commands parameter use a default value.
this means existing users of the command do not need to specify it in the command script

digital_land/package/dataset.py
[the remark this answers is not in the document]
morgan-sl: agreed to park this for now.
the code has 2 pathways which both need to make use of the functionality in insert_entity.
expanding it out would mean repeating sections of code.
a new Trello ticket will be created to capture that this needs revisiting.

digital_land/package/dataset.py
```diff
@@ -119,15 +120,23 @@ def entity_row(self, facts):
...
         row = self.migrate_entity(row)
         if row:
-            self.insert("entity", self.entity_fields, row)
+            entity_fields = [
+                field for field in self.entity_fields if not field.endswith("-geom")
+            ]
+            self.add_table_values(entity_fields, row, insert_rows)
+            if len(insert_rows) >= self.max_batch_size:
+                self.insert_many("entity", self.entity_fields, insert_rows)
+                self.commit()
```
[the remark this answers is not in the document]
morgan-sl: It doesn't currently, but there's no reason not to.
I'll add it in Revision V

digital_land/cli.py
```diff
@@ -108,17 +108,21 @@ def convert_cmd(input_path, output_path):
 
 @cli.command("dataset-create", short_help="create a dataset from processed resources")
 @click.option("--output-path", type=click.Path(), default=None, help="sqlite3 path")
+@click.option("--max-batch-size", type=click.INT)
```
eveleighoj, line 111: This will break the pipeline, there is no default it could be worth making an acceptance test for creating a dataset

eveleighoj (approved):

PR #149, opened 2023-11-03
Convert phase: Adds saving converted input as a csv

CarlosCoelhoSL:
This PR adds functionality to the convert phase so that when the input file (geojson, xlsm etc) is converted to a .csv file, this is saved to disk (rather than a temporary file). The location of where the file is saved is determined by arguments for the ConvertPhase class. If these arguments aren't present then the file isn't saved to disk.

This is required so the pipeline-runner can use the converted input .csv as part of its response.

### DIFF
--- a/digital_land/phase/convert.py
+++ b/digital_land/phase/convert.py
@@ -114,16 +114,29 @@ def convert_features_to_csv(input_path):
 
 
 class ConvertPhase(Phase):
-    def __init__(self, path=None, dataset_resource_log=None, custom_temp_dir=None):
+    def __init__(
+        self,
+        path=None,
+        dataset_resource_log=None,
+        custom_temp_dir=None,
+        custom_temp_file=None,
+    ):
         self.path = path
         self.log = dataset_resource_log
         self.charset = ""
         # Allows for custom temporary directory to be specified
         # This allows symlink creation in case of /tmp & path being on different partitions
         if custom_temp_dir:
+            self.custom_temp_dir = custom_temp_dir
             self.temp_file_extra_kwargs = {"dir": custom_temp_dir}
+            if not os.path.exists(custom_temp_dir):
+                os.makedirs(custom_temp_dir)
         else:
             self.temp_file_extra_kwargs = {}
+        # custom_temp_file
+        # allows a specified file if user wants to interrogate them
+        if custom_temp_file:
+            self.custom_temp_file = custom_temp_file
 
     def process(self, stream=None):
         input_path = self.path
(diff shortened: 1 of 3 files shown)


tests/unit/test_convert.py
```diff
@@ -13,3 +13,17 @@ def test_load_xlsm():
     assert block["line-number"] == 1
     assert "OrganisationURI" in block["line"]
     assert log.mime_type == "application/vnd.ms-excel"
+
+
+def test_converted_csv_saved():
```
eveleighoj, line 18: This test only covers .xlsm, if you tried .geojson or .gpkg. This only works when a reader exists. It will also copy a csv from one place to another when we don't need it to.

digital_land/phase/convert.py
```diff
@@ -114,16 +114,29 @@ def convert_features_to_csv(input_path):
 
 
 class ConvertPhase(Phase):
-    def __init__(self, path=None, dataset_resource_log=None, custom_temp_dir=None):
+    def __init__(
+        self,
+        path=None,
+        dataset_resource_log=None,
+        custom_temp_dir=None,
+        custom_temp_file=None,
```
eveleighoj, line 122: I would be tempted not to use custom_temp_file for now and just name the file the same as the input file (I know I used custom_temp_file in mine but there should be a way to not use it)

eveleighoj: Just a warning on the above, the custom_temp_dir is used for something so could be worth double checking what it's used for right now. I haven't spent too long looking at the tempfile package.

digital_land/phase/convert.py
```diff
@@ -143,6 +156,15 @@ def process(self, stream=None):
             # raise StopIteration()
             reader = iter(())
 
+        # save converted input to a file if custom_temp_dir and custom_temp_file are set
+        if hasattr(self, "custom_temp_dir") and hasattr(self, "custom_temp_file"):
```
eveleighoj, line 160: I would edit the ogr2ogr command (convert_features_to_csv) rather than something like this as otherwise you might end up making an extra copy of data which could impact large datasets a lot.

eveleighoj (commented):
Needs to be more throughly testing with the core file types we ask for from LPAs. Approach should be change to give a different file name in the convert_features_to_csv function.

tests/unit/test_convert.py
```diff
@@ -13,3 +13,55 @@ def test_load_xlsm():
     assert block["line-number"] == 1
     assert "OrganisationURI" in block["line"]
     assert log.mime_type == "application/vnd.ms-excel"
+
```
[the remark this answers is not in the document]
eveleighoj: Another test to write would be one that could test for data not being in the converted folder and instead being in a user data

PR #152, opened 2023-11-15
Shapely upgrade

kena-SL:
Card: https://trello.com/c/tRAHx5qd/1059-upgrade-to-shapely-201

Description:
This PR **upgrades Shapely** to version 2.0.2
**Removes loop** in wkt.py file
uses **set_precision()** function to define fixed grid size
Adds **validity check** at the end of processing and raises issue accordingly
Records and fixes issue raised by **make_multipolygon()** function

Testing:
test added to verify invalid geometry getting fixed and asserts if the difference in area between provided and final geometry is greater than 95

Moved tests from tests/unit/test_wkt.py to tests/unit/datatype/test_wkt.py


digital_land/datatype/wkt.py
```diff
@@ -166,6 +169,11 @@ def normalise_geometry(geometry, simplification=0.000005):
     # ensure geometry is a MultiPolygon
     geometry = make_multipolygon(geometry)
 
+    if geometry:
+        if not geometry.is_valid:
+            issue = explain_validity(geometry)
```
eveleighoj, line 174: I'm assuming this is needed incase the make_polygon function creates an invalid geometry?

In which case I would suggest supplying the original issue instead of this issue

kena-SL: Yes, to fix issues introduced by make_multipolygon

Function updated to return original issue👍

tests/unit/datatype/test_wkt.py
```diff
@@ -82,16 +84,15 @@ def test_normalise_geometrycollection_provided():
...
-        "MULTIPOLYGON (((-0.173644 51.549807,-0.173551 51.549811,-0.173545 51.549902,"
-        "-0.173545 51.549912,-0.173554 51.549913,-0.173550 51.549971,-0.173689 51.549976,"
-        "-0.173724 51.549935,-0.174041 51.549928,-0.174039 51.549828,-0.174024 51.549799,"
-        "-0.173644 51.549807)),"
-        "((-0.176790 51.555154,-0.176884 51.555212,-0.176896 51.555199,-0.176994 51.555231,"
-        "-0.177025 51.555233,-0.177041 51.555212,-0.177085 51.555194,-0.176995 51.555106,"
-        "-0.176900 51.555147,-0.176851 51.555112,-0.176790 51.555154)),"
-        "((-0.176757 51.552538,-0.176804 51.552519,-0.176793 51.552502,-0.176758 51.552515,"
-        "-0.176707 51.552463,-0.176595 51.552506,-0.176651 51.552562,-0.176746 51.552526,"
-        "-0.176757 51.552538)))"
```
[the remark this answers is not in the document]
eveleighoj: Just to jump on this, with all the tests below here it's hard for me to tell what's changed. But if your changing the hard coded numbers for new ones I think the tests needs to altered as they aren't performing how they should be.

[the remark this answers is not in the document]
kena-SL: Moved input wkts from these tests (test_wkt_invalid_multipolygon_self_intersection_with_holes() and test_invalid_geometry_fixed_and_area_maintained()) to the new parameterized test

digital_land/datatype/wkt.py
```diff
@@ -166,6 +169,10 @@ def normalise_geometry(geometry, simplification=0.000005):
     # ensure geometry is a MultiPolygon
     geometry = make_multipolygon(geometry)
 
+    if geometry:
+        if not geometry.is_valid:
+            geometry = geometry.buffer(0)
```
eveleighoj, line 174: What is the purpose of this?

kena-SL: make_multipolygon() function does introduce issues like self-intersection in a valid geometry
Buffer attempts to fix any invalidities in the resulting MultiPolygon similar to make_valid()

eveleighoj (approved):

eveleighoj (changes requested):
This currently fails if there is no issue log  (i.e. issues=None) this is a problem as it means the arguement is no longer optional.

Can we add a test that runs with no issue log and make sure it wokrs?

eveleighoj (approved):

PR #155, opened 2023-11-29
Adding a null check

ssadhu-sl:
Adding a null check so that severity and description are not appended when issue log is empty


CarlosCoelhoSL (approved):
lgtm

PR #158, opened 2023-12-22
Appending Error Messages in pipeline runner API

ssadhu-sl:
Ticket: https://trello.com/c/dYjAIi79/1120-mvp-make-the-issues-meaningful-on-the-error-page

Appending error message description dynamically from issue log so it contains which field name is that error for specifically. This is specifically included for pipeline runner and will be called only when triggered from there.

### DIFF
--- a/tests/unit/test_log.py
+++ b/tests/unit/test_log.py
@@ -1,20 +1,42 @@
 import pytest
 from digital_land.log import IssueLog
-
-test_collection_dir = "tests/data"
-
-
-@pytest.fixture
-def sample_issue_log_csv_path():
-    return test_collection_dir + "/specification/issue-type.csv"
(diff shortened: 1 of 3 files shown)


tests/unit/test_log.py
```diff
@@ -9,6 +9,11 @@ def sample_issue_log_csv_path():
     return test_collection_dir + "/specification/issue-type.csv"
 
 
+@pytest.fixture
```
eveleighoj, line 12: These tests rely on a file system hence should be in integration tests

eveleighoj (changes requested):

eveleighoj (approved):
Looks good, there is a mocker fixture that you can use with pytest but I don't think it really matters as long as it's mocked!

PR #160, opened 2024-01-02
Improve add endpoint and lookups

ssadhu-sl:
Ticket: https://trello.com/c/5eWH3OOY/547-improve-add-endpoints-process-for-adding-lookups-when-there-are-no-existing-lookups

Improving the add lookup process by ensuring the entity is within the entity defined in specification for the dataset

### DIFF
--- a/digital_land/pipeline.py
+++ b/digital_land/pipeline.py
@@ -400,14 +400,14 @@ class Lookups:
 
         if len(self.entries) > 0:
             # check entry is being run against the correct dataset
-            existing_prefixes = len(
-                [1 for item in self.entries if item["prefix"] == entry["prefix"]]
-            )
+            # existing_prefixes = len(
+            #     [1 for item in self.entries if item["prefix"] == entry["prefix"]]
+            # )
 
-            if existing_prefixes == 0:
-                raise ValueError(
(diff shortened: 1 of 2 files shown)


digital_land/pipeline.py
```diff
@@ -400,14 +400,14 @@ def validate_entry(self, entry) -> bool:
 
         if len(self.entries) > 0:
             # check entry is being run against the correct dataset
-            existing_prefixes = len(
```
eveleighoj, line 403: I would delete this code

tests/e2e/test_add_endpoints_and_lookups.py
```diff
@@ -248,60 +248,6 @@ def test_command_add_endpoints_and_lookups_success_lookups_required(
         assert endpoint.get("entry-date", None) == expected_entry_date
 
 
-def test_command_add_endpoints_and_lookups_failure_incorrect_dataset(
```
eveleighoj, line 251: Yes right to remove it, but I'm wondering with the above removal of an error and this test is there any other errors that can occur e.g. range max has been hit or maybe a warning that states the current is below the minimum?

eveleighoj: They may already be included

eveleighoj (approved):

PR #161, opened 2024-01-15
(re)add entry-number to lookup.csv.

cjohns-scottlogic:
…v file is read which has extra columns.


CarlosCoelhoSL (approved):
sad we won't be accepting wensleydayle cheese but lgtm!

PR #163, opened 2024-01-19
Update add endpoints and lookups process to consider entity numbers in old-entity.csv

kena-SL:
Card: https://trello.com/c/lFFK6cJl/977-update-process-for-add-endpoint-to-consider-old-entity-numbers

This PR updates the save_csv() function in the Lookups class
Before assigning new entity numbers to unassigned lookups it checked for those numbers in old-entity.csv file

If the entity number exists already in old-entity.csv (for any reason) then it is avoided and not assigned

This is to prevent any invalid redirections for any new entities added

**Test**
the test sets up a mock CSV file path and its content 
uses the patch to mock the open function
the test ensures that the new lookup doesn't get assigned a entity number that exists in the old-entity file

### DIFF
--- a/digital_land/commands.py
+++ b/digital_land/commands.py
@@ -515,11 +515,6 @@ def add_endpoints_and_lookups(
             writer = csv.writer(f)
             writer.writerow(list(lookups.schema.fieldnames))
 
-    if not os.path.exists(lookups.old_entity_path):
-        with open(lookups.old_entity_path, "w", newline="") as f:
-            writer = csv.writer(f)
-            writer.writerow(["old-entity", "status", "entity"])
-
     lookups.load_csv()
     for new_lookup in new_lookups:
         for idx, entry in enumerate(new_lookup):
--- a/digital_land/commands.py
+++ b/digital_land/commands.py
@@ -517,6 +517,11 @@ def add_endpoints_and_lookups(
                 ["prefix", "resource", "organisation", "reference", "entity"]
             )
 
+    if not os.path.exists(lookups.old_entity_path):
+        with open(lookups.old_entity_path, "w", newline="") as f:
+            writer = csv.writer(f)
+            writer.writerow(["old-entity", "status", "entity"])
+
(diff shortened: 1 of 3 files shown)


digital_land/pipeline.py
```diff
@@ -389,12 +390,25 @@ def save_csv(self, lookups_path=None, entries=None):
             f, fieldnames=self.schema.fieldnames, extrasaction="ignore"
         )
         writer.writeheader()
+
+        old_entity_path = self.old_entity_path
+        reader = csv.DictReader(open(old_entity_path, newline=""))
```
cjohns-scottlogic, line 395: Should probably handle the case where old_entity_path doesn't exist.

digital_land/commands.py
```diff
@@ -517,6 +517,11 @@ def add_endpoints_and_lookups(
                 ["prefix", "resource", "organisation", "reference", "entity"]
             )
 
+    if not os.path.exists(lookups.old_entity_path):
+        with open(lookups.old_entity_path, "w", newline="") as f:
+            writer = csv.writer(f)
+            writer.writerow(["old-entity", "status", "entity"])
+
```
cjohns-scottlogic, line 524: Do we need to add an empty old entity file if we can handle it not being there?

digital_land/pipeline.py
```diff
@@ -389,12 +390,26 @@ def save_csv(self, lookups_path=None, entries=None):
             f, fieldnames=self.schema.fieldnames, extrasaction="ignore"
         )
         writer.writeheader()
+
+        entity_values = []
+        if os.path.exists(self.old_entity_path):
```
cjohns-scottlogic, line 395: This seems to be tripping up the test. I'd suggest either mocking it so the test thinks the file exists, or create the DictReader in a try .. except block and silently ignore a FileNotFoundError (but maybe re-raise any other sort)

cjohns-scottlogic (approved):

PR #164, opened 2024-01-19
Fixes entity ranges in the add-endpoint-and-lookups script

CarlosCoelhoSL:
This PR fixes a problem in which the add-endpoints-and-lookups script would sometimes fetch incorrect dataset entity min/max values. It also changes the lookup.csv generation when there is no file detected to use the schema rather than a list of hard coded columns.

This problem happened due to the `dataset` variable from a prior loop (with an unpredictable value) mistakenly being used to fetch the dataset entity min/maxes.

Ticket: https://trello.com/c/Et5FCInU/993-developer-contributions-fix-add-endpoints-and-lookups-behaviour

### DIFF
--- a/digital_land/commands.py
+++ b/digital_land/commands.py
@@ -513,9 +513,7 @@ def add_endpoints_and_lookups(
     if not os.path.exists(lookups.lookups_path):
         with open(lookups.lookups_path, "w", newline="") as f:
             writer = csv.writer(f)
-            writer.writerow(
-                ["prefix", "resource", "organisation", "reference", "entity"]
-            )
+            writer.writerow(list(lookups.schema.fieldnames))
 
     lookups.load_csv()
     for new_lookup in new_lookups:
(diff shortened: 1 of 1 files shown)


digital_land/commands.py
```diff
@@ -513,9 +513,7 @@ def add_endpoints_and_lookups(
     if not os.path.exists(lookups.lookups_path):
         with open(lookups.lookups_path, "w", newline="") as f:
             writer = csv.writer(f)
-            writer.writerow(
-                ["prefix", "resource", "organisation", "reference", "entity"]
-            )
+            writer.writerow(list(lookups.schema.fieldnames))
```
ssadhu-sl, line 516: It will pull all the fields from the lookup schema. Are we okay with that?

CarlosCoelhoSL: It should just pull the fields we expect to see in lookup files i.e `['prefix', 'resource', 'entry-number', 'organisation', 'reference', 'entity']`

ssadhu-sl: Interesting. Doesn't it use the specification schema lookup file to pull it from. Not sure as I haven't seen the full code. Just curious if it takes from here or somewhere else. (https://github.com/digital-land/specification/blob/main/content/dataset/lookup.md)

CarlosCoelhoSL: It takes it from [schema.py]((https://github.com/digital-land/digital-land-python/blob/main/digital_land/schema.py) in digital-land-python.
[This ticket](https://trello.com/c/AzdxYqBL/988-deal-with-additional-column-in-lookup-file-compared-to-brownfield-land-add-endpoints-and-lookups-script) that is already completed enforces that the columns in lookup.csv match the columns for lookup schema.py. 
I just realised that the file creation wasn't following these same columns so changed them to match

ssadhu-sl (approved):
lgtm

PR #165, opened 2024-01-29
Add missing value

ssadhu-sl:
Reference Ticket: https://trello.com/c/j8fhkVhg/1158-file-with-empty-geometry-being-processed-successfully

- Adding a new issue-type 'Missing Value' for empty geometries for Geometry typologies.
- Also, changed the map phase to only take the row keys as the ones given by LPA in their data and not all the fields from specification. 
- Changed some test cases to reflect above changes


kena-SL (approved):
Looks good!

PR #172, opened 2024-02-12
Add entity redirects

kena-SL:
Ref: https://trello.com/c/kHTiSrmx/999-merge-existing-geographically-duplicated-conservation-area-entities

This PR creates a command "add-redirections" which takes a csv file (which contains list of duplicates entities) and assign redirects for those to the old-entity.csv (to merge the entities)

### DIFF
--- a/digital_land/commands.py
+++ b/digital_land/commands.py
@@ -746,3 +746,48 @@ def process_data_in_batches(entities, flattened_dir, dataset_name):
         )
 
     return feature_collection
+
+
+def add_redirections(csv_file_path, pipeline_dir):
+    """
+    :param csv_file_path:
+    :param pipeline_dir:
+    :return:
+    """
+    expected_cols = [
+        "entity_HE",
+        "entity_LPA",
+    ]
+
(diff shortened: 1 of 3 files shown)


digital_land/commands.py
```diff
@@ -746,3 +746,48 @@ def process_data_in_batches(entities, flattened_dir, dataset_name):
...
+def add_redirections(csv_file_path, pipeline_dir):
+    """
+    :param csv_file_path:
+    :param pipeline_dir:
+    :return:
+    """
+    expected_cols = [
+        "entity_HE",
+        "entity_LPA",
+    ]
```
cjohns-scottlogic, line 760: I presume these column names are for Historic England and the LPA? Could/should we use a more general name in case this process needs to be done again for something else? I'll take @eveleighoj here as he might have some input on that.

kena-SL: Yes that makes sense!
Can we have something like "entity_source " and "entity_destination" ?

cjohns-scottlogic (approved):

PR #175, opened 2024-02-20
Add mandatory fields

JbannisterScottLogic:
Have documentation-url and licence as mandatory fields and check specification for licence


kena-SL (commented):
License check looks good.
For documentation-url, It checks for the field present in CSV but I think it will accept the field with empty value

kena-SL (approved):

PR #182, opened 2024-03-05
Adds dataset to stream

CarlosCoelhoSL:
This PR fixes a bug found on [this ticket](https://trello.com/c/M927WA1j/1261-fix-dataset-in-stream-in-harmonise-phase) where the dataset is not being correctly set in the stream that is passed through the phases. This has been added to the convert phase.

EDIT: After discussing with Owen we have decided to pass the dataset directly to the harmonise phase rather than through the stream

Also includes minor improvements to the missing value mandatory fields code and tests.


JbannisterScottLogic (approved):
lgtm

PR #183, opened 2024-03-07
Replace None with an empty string during map phase.

cjohns-scottlogic:
This reinstates the previous behavior of the map phase when a column was not present.


CarlosCoelhoSL (approved):
lgtm

CarlosCoelhoSL (approved):
lgtm

PR #184, opened 2024-03-22
Catch all RequestExceptions so we don't crash for errors from requests.

cjohns-scottlogic:


digital_land/collect.py
```diff
@@ -77,15 +77,7 @@ def get(self, url, log={}, verify_ssl=True, plugin="get"):
...
-        except (
-            requests.exceptions.SSLError,
-            requests.ConnectionError,
-            requests.HTTPError,
-            requests.Timeout,
-            requests.TooManyRedirects,
-            requests.exceptions.MissingSchema,
-            requests.exceptions.ChunkedEncodingError,
-        ) as exception:
+        except (requests.RequestException,) as exception:
```
eveleighoj, line 80: Does this work for all of the current errors?

tests/unit/test_collect.py
```diff
@@ -91,10 +91,17 @@ def log_file(url):
 
 
 @responses.activate
-def test_get(collector, prepared_exception):
+@pytest.mark.parametrize(
+    "ex_cls,msg",
+    [
+        (requests.ConnectionError, "Test Connection Error"),
+        (requests.exceptions.ContentDecodingError, "Test Content Decoding Error"),
```
eveleighoj, line 98: Is it worth adding all the known ones above?

digital_land/collect.py
cjohns-scottlogic: requests.RequestException seems to be a base of all the others, and the tests seem prove that.

eveleighoj (approved):

PR #185, opened 2024-03-26
remove spec from phases and add tests

eveleighoj:
We need to do some refactoring of the code base. Right now the commands file is huge and will only keep growing unless we break down commands and processes into their own modules. The issue with the pipeline commands right now is that they are reliant on connecting directly to a specification object. This makes it hard to test as you always need to use the specification when actually outputs should just be shared.

This is the first step in separating this out. Allowing us to test just the transformations. This will help separate the functions for the possibility of moving them to their own module in the project.


cjohns-scottlogic (approved):

ssadhu-sl (dismissed):
looks good!

PR #188, opened 2024-04-09
Post conversion expectations

JbannisterScottLogic:
Expectations on converted resources

### DIFF
--- a/digital_land/expectations/checkpoints/converted_resource.py
+++ b/digital_land/expectations/checkpoints/converted_resource.py
@@ -5,9 +5,75 @@
 # a checkpoint represents the moment in the process where we tell it the
 # type of data it is validating and where the data is
 # the primary different between checkpoints is how it loads expectations (i.e. where that are loaded from)
+from pathlib import Path
+import csv
+import re
 from .base import BaseCheckpoint
 
 
-class CovertedResourceCheckpoint(BaseCheckpoint):
-    def load():
-        pass
+class ConvertedResourceCheckpoint(BaseCheckpoint):
+    def __init__(self, data_path):
+        super().__init__("converted_resource", data_path)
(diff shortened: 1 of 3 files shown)


digital_land/expectations/issue.py
```diff
@@ -19,6 +19,8 @@ def issue_factory(scope):
         "row-group": RowGroupIssue,
         "row": RowIssue,
         "value": ValueIssue,
+        "duplicate_reference": DuplicateReferenceIssue,
```
cjohns-scottlogic, line 22: I'm not sure if this should be a "scope" or not? I thought scope was for the "level" of what's wrong, so it would be a "row" scope, and the DuplicateReferenceIssue and InvalidReferenceIssue classes inherit from RowIssue. I'll tag @eveleighoj for his input.

eveleighoj: No I agree, the scope of an issue should affect what details need to be provided. the scope I think for this should be a row if the issue is specific to a row or a row group if it's multiple rows.

digital_land/expectations/checkpoints/converted_resource.py
```diff
@@ -5,9 +5,82 @@
...
-        pass
+class ConvertedResourceCheckpoint(BaseCheckpoint):
+    def __init__(self, data_path):
+        super().__init__("converted_resource", data_path)
+        self.csv_path = Path(data_path)
+
+    def load(self):
+        self.expectations = [
+            {
+                "function": self.check_for_duplicate_references,
```
eveleighoj, line 21: The function should be in the expectation_functions folder instead of a method on the class. Also see the loading function for the dataset checkpoint it dates three different lists in depending on the dataset and the dataset typology

PR #192, opened 2024-04-24
assign functionality to lookupPhase

kena-SL:


digital_land/phase/lookup.py
```diff
@@ -59,6 +68,9 @@ def process(self, stream):
                         # or by the CURIE
                         or self.lookup(prefix=prefix, reference=reference)
                     )
+                if self.redirect_lookups:
```
eveleighoj, line 71: We need to factorise across LookupPhase and PrintLookupPhase to stop them separating in the future. currently they're using different code to do the same thing which they shouldn't be. Specifically PrintLookupPhase could be another child of the Lookup phase class

eveleighoj:
Finally, can you check the impact of logging unknown entity errors, we need to make sure we're not logging them as missing when we're removing them.

eveleighoj (changes requested):

digital_land/phase/prune.py
```diff
@@ -30,8 +30,7 @@ class EntityPrunePhase(Phase):
     remove entries with a missing entity
     """
 
-    def __init__(self, issue_log=None, dataset_resource_log=None):
-        self.issues = issue_log
+    def __init__(self, dataset_resource_log=None):
```
eveleighoj, line 33: leave issue_log in for now but raise a depreciation warning otherwise it may break Samriti's API code

digital_land/phase/lookup.py
```diff
@@ -24,12 +24,24 @@ class LookupPhase(Phase):
...
+
+    def __init__(self, lookups={}, redirect_lookups={}, issue_log=None):
         self.lookups = lookups
+        self.redirect_lookups = redirect_lookups
+        self.issues = issue_log
 
     def lookup(self, **kwargs):
         return self.lookups.get(key(**kwargs), "")
 
+    def get_entity_number(self, entity):
```
eveleighoj, line 37: this name is a bit misleading, the function is checking for a redirect and returning the new entity number if it it (or none if it's been retired).

you could expand the function even more and include getting the number from lookups as well then re use it across both classes rather than needing to use super

eveleighoj (changes requested):

eveleighoj (approved):

PR #196, opened 2024-05-17
Add retire-endpoints-and-sources script

amiriam-scottlogic:
Created a script which should retire endpoints and sources when given a csv of endpoints/sources with the columns: "collection", "endpoint" and "source".


digital_land/collection.py
```diff
@@ -555,3 +557,45 @@ def entry_date(self, entry):
         return entry.get(
             "entry-date", datetime.utcnow().strftime("%Y-%m-%dT%H:%H:%M:%SZ")
         )
+
+    def retire_endpoints_and_sources(collection, collection_df_to_retire) -> None:
```
JbannisterScottLogic, line 561: Great use of a separate function to handle the retirement logic

digital_land/collection.py
```diff
@@ -555,3 +557,45 @@ def entry_date(self, entry):
...
+                source_csv_df["source"].isin(collection_df_to_retire["source"]),
+                "end-date",
+            ] = today_date
+
+            # Write updated data back to CSV files
+            endpoint_csv_df.to_csv(endpoint_csv_path, index=False)
+            source_csv_df.to_csv(source_csv_path, index=False)
+
+        except FileNotFoundError as e:
+            print(f"Error: {e}. Please check the file paths and try again.")
```
JbannisterScottLogic, line 599: Consider using logging instead of print statements for better error tracking

digital_land/collection.py
```diff
@@ -555,3 +557,45 @@ def entry_date(self, entry):
...
+            collection_df_to_retire (str): A small dataframe of sources and endpoints to be retired within a colleciton, this should be
+                generated by a function call within commands.py.
+        """
+
+        try:
+            endpoint_csv_path = os.path.join(collection.dir, "endpoint.csv")
+            source_csv_path = os.path.join(collection.dir, "source.csv")
+
+            # Read endpoint and source CSV files
+            endpoint_csv_df = pd.read_csv(endpoint_csv_path)
```
JbannisterScottLogic, line 576: Consider adding a check to ensure the CSV files exist before attempting to read them

tests/unit/test_collection.py
```diff
@@ -90,3 +94,59 @@ def test_format_date():
...
+    expected_endpoint_data = {
+        "endpoint": {0: "endpoint1", 1: "endpoint2", 2: "endpoint3"},
+        "end-date": {0: today_date, 1: np.nan, 2: today_date},
+    }
+    expected_source_data = {
+        "source": {0: "source1", 1: "source2", 2: "source3"},
+        "end-date": {0: np.nan, 1: today_date, 2: today_date},
+    }
+
+    assert updated_endpoint_df.to_dict() == expected_endpoint_data
```
JbannisterScottLogic, line 151: These assertions effectively verify that the end-date columns are updated correctly. Consider adding more assertions to test edge cases where no updates are needed

digital_land/collection.py
```diff
@@ -555,3 +557,45 @@ def entry_date(self, entry):
...
+            ] = today_date
+
+            # Update end_date column in source file
+            source_csv_df.loc[
+                source_csv_df["source"].isin(collection_df_to_retire["source"]),
+                "end-date",
+            ] = today_date
+
+            # Write updated data back to CSV files
+            endpoint_csv_df.to_csv(endpoint_csv_path, index=False)
```
JbannisterScottLogic, line 595: Good job ensuring the updated data is written back to the CSV files. For large files, consider ways to optimize this process to only write necessary updates.

JbannisterScottLogic (commented):
Looks good to me! Just a few comments for consideration

digital_land/collection.py
amiriam-scottlogic: The try/catch checks for FileNotFoundErrors so we should be okay

digital_land/collection.py
```diff
@@ -555,3 +557,45 @@ def entry_date(self, entry):
...
+            endpoint_csv_df.loc[
+                endpoint_csv_df["endpoint"].isin(collection_df_to_retire["endpoint"]),
+                "end-date",
+            ] = today_date
+
+            # Update end_date column in source file
+            source_csv_df.loc[
+                source_csv_df["source"].isin(collection_df_to_retire["source"]),
+                "end-date",
+            ] = today_date
```
CarlosCoelhoSL, line 592: Would it be possible to display the sources/endpoints that have been ended? Or even just log an error for the sources/endpoints that weren't found in the endpoint.csv and source.csv files? As someone may run this with a whole batch of sources, one or two of them have a mistake in the hash and not found but they'd currently have no idea

digital_land/collection.py
amiriam-scottlogic: To achieve this my only idea is to create custom csv write logic to preserve the rows that I don't change within the function, This might defeat the purpose of optimization.

digital_land/collection.py
amiriam-scottlogic: Logging is definitely a better method for error handling but I'm not sure logging is necessary for this simple of a script - although I could be wrong

digital_land/collection.py
amiriam-scottlogic: This is a very good point, I'll look to modifying this

amiriam-scottlogic: @CarlosCoelhoSL I've adjusted the function to fail completely if any source/endpoint hashes have no match found, and then print these values for reference

JbannisterScottLogic (dismissed):

JbannisterScottLogic (approved):

PR #198, opened 2024-05-23
Add command to create organisations.csv

cjohns-scottlogic:


digital_land/package/organisation.py
```diff
@@ -0,0 +1,266 @@
...
+import csv
+import sys
+import json
+import requests
+import logging
+from pathlib import Path
+
+from .csv import CsvPackage
+
+DATASET_URL = "https://files.planning.data.gov.uk/organisation-collection/dataset/"
```
ssadhu-sl, line 11: Should we define it here? what do you think about putting it in a makefile or some centralized place?

ssadhu-sl (approved):
lgtm

PR #199, opened 2024-05-23
Alter filters for endpoints selection

JbannisterScottLogic:
Creating new filters to allow for pipeline to filter by endpoint


digital_land/pipeline.py
```diff
@@ -194,13 +203,20 @@ def load_redirect_lookup(self):
             old_entity = row.get("old-entity", "")
             entity = row.get("entity", "")
             status = row.get("status", "")
-            if old_entity and status:
+            if old_entity and entity and status:
```
eveleighoj, line 206: is this on purpose? I think if the status is 410 then there doesn't need to be an entity

digital_land/pipeline.py
```diff
@@ -194,13 +203,20 @@ def load_redirect_lookup(self):
...
                 self.redirect_lookup[old_entity] = {"entity": entity, "status": status}
 
-    def filters(self, resource=""):
-        d = self.filter.get("", {})
+    def filters(self, resource="", endpoints=""):
```
eveleighoj: the default endpoint here is a string but should it be a list?

digital_land/pipeline.py
JbannisterScottLogic: Had some merge issues and was added in error. Corrected now

digital_land/pipeline.py
JbannisterScottLogic: Line from an older commit. Filters now takes the list of endpoints accordingly

CarlosCoelhoSL (approved):
lgtm

PR #200, opened 2024-05-29
Filter on column

ssadhu-sl:
Reference Ticket: https://trello.com/c/kUQ0fYdl/3189-ensure-filters-can-apply-to-columns-and-not-just-field

This enhancement includes:
   - Filter Phase is now called before Map phase as well to make sure filter is done on any column of data and not just mapped fields/fields in specifications
   - Integration test to make sure the filter is applied on column of LPA data which is not in specification or mapped to any field 
      in specification


cjohns-scottlogic (commented):
lgtm

cjohns-scottlogic (approved):

PR #202, opened 2024-06-06
Assign-entities Bug fix

ssadhu-sl:
Reference tickets: 
- https://trello.com/c/ZHyP2EMv/3214-fix-assign-entities-command-to-work-with-config-repository
- https://trello.com/c/WhiaJS0B/3231-add-endpoints-and-lookups-script-bug

This ticket includes the following changes:
1. Fix assign-entities to make sure it runs on config on it's own without needing collection folders
2. Ensure the column mapping is done via both resource and endpoint hash
3. Add relevant test cases

### DIFF
--- a/tests/e2e/test_assign_entities.py
+++ b/tests/e2e/test_assign_entities.py
@@ -212,16 +212,19 @@ def test_command_assign_entities(
     """
     collection_name = "ancient-woodland"
     dataset_name = "ancient-woodland"
+    test_endpoint = "d779ad1c91c5a46e2d4ace4d5446d7d7f81df1ed058f882121070574697a5412"
 
     collection = Collection(name=collection_name, directory=collection_dir)
-    collection.load()
 
     assign_entities(
         resource_file_paths=["mock_csv.csv"],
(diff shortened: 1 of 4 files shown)


tests/e2e/test_assign_entities.py
```diff
@@ -328,64 +344,24 @@ def test_command_assign_entities_no_reference_log(
...
         pipeline_dir=pipeline_dir,
+        dataset=dataset,
+        endpoints=test_endpoint,
     )
-
     assert "No reference" in caplog.text
     assert "mock_csv" in caplog.text
 
 
-def test_command_assign_entities_resource_not_processed(
```
cjohns-scottlogic, line 346: Just checking this test is no longer required.

ssadhu-sl: No we have isolated assign_entities now and we will provide the resource path, if not valid it will throw No Resource found which is included in other test case. This test case no longer is valid

cjohns-scottlogic (approved):

PR #203, opened 2024-06-06
Handle "efficient json" format

cjohns-scottlogic:
For json files, we check if the file is in 'efficient json' format we convert it ourselves. Otherwise we call out to ogr2ogr as before.

### DIFF
--- a/digital_land/phase/convert.py
+++ b/digital_land/phase/convert.py
@@ -115,24 +115,35 @@ def convert_features_to_csv(input_path, output_path=None):
     return output_path
 
 
+def save_efficient_json_as_csv(output_path, columns, data):
+    with open(output_path, "w") as csv_file:
+        cw = csv.writer(csv_file)
+        cw.writerow(columns)
+
+        for row in data:
+            cw.writerow(row)
+
+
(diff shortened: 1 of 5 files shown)


digital_land/phase/convert.py
```diff
@@ -114,6 +115,28 @@ def convert_features_to_csv(input_path, output_path=None):
...
+            if item[0] in ["columns"]:
+                columns = [x for x in item[1].persistent()]
+
+            if item[0] in ["data"] and columns is not None:
+                with open(output_path, "w") as csv_file:
+                    cw = csv.writer(csv_file)
+                    cw.writerow(columns)
+
+                    for row in item[1]:
+                        cw.writerow(row)
```
CarlosCoelhoSL, line 133: Does this work if data is before columns in the file?

cjohns-scottlogic: I suspect not, but if it's not we can't process it as a stream.

cjohns-scottlogic: Updated to handle the data being first. It won't stream it in this case, but it will handle it.

CarlosCoelhoSL (approved):
lgtm!

PR #206, opened 2024-06-24
Adds output path generation if none is supplied to convert phase (efficient json format)

CarlosCoelhoSL:
In the convert phase when handling the 'efficient' json format, if no output path was given to the phase then no conversion would happen. This PR sets the output path to a temporary file if none is supplied


ssadhu-sl (approved):
lgtm

PR #207, opened 2024-06-25
Disable size limit on later GDAL versions

eveleighoj:
The border collections action fails because each individual geometry is so massive ogr2ogr has an environment variable which can tackle this but only for more recent version. This PR adds it in. but needs more work so that it only adds it in for recent versions

the test. runner will need updating with a more recent version of GDAL too, above 3.5.2

see ticket [here](https://trello.com/c/1Q3Hkfp5/3275-migrate-gdal-versions-in-the-github-runner)


digital_land/phase/convert.py
```diff
@@ -4,16 +4,28 @@
...
 import tempfile
 import zipfile
+from packaging.version import Version
 from io import StringIO
 import pandas as pd
 from .load import Stream
 from .phase import Phase
 
 
+def get_gdal_version():
```
cjohns-scottlogic, line 19: @eveleighoj Is there a suitable place to re-factor both copies of this function to?

cjohns-scottlogic (approved):

PR #214, opened 2024-07-11
Priotise endpoint/resource column mappings

kena-SL:
Given multiple columns that map to same field endpoint or resource specific column mappings should take priority.

e.g for these columns mappings where two different columns are mapped to the same field.

```
pipeline,resource,column,field
pipeline-one,,quatre,four
pipeline-one,some-resource,quatro,four
```
The resulting mappings seems to contain both of them.
This PR updates this such that the resulting mapping has only one resource specific column mapping in it.

The other general mappings remain the same.

### DIFF
--- a/digital_land/pipeline.py
+++ b/digital_land/pipeline.py
@@ -234,6 +234,11 @@ class Pipeline:
         for key in general_columns:
             if key in result:
                 continue
+            if (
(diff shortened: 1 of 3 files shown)


ssadhu-sl (approved):
lgtm

PR #215, opened 2024-07-11
More tests for organisation package.

cjohns-scottlogic:
Added tests for files not part of the org package, and for typology when generating organisaion.csv


CarlosCoelhoSL (approved):
lgtm

PR #218, opened 2024-07-15
Update load lookups

kena-SL:
Currently, the pipeline assigns entities based on prefix and reference matching without considering organisation. This can result in instances where different organisations may share common entities when their references match.

When references match across organizations, the command currently fails to treat the entities of the second organization as new.

The reason being load_lookups() function includes entries that only have a prefix and reference. 
This PR updates the function to add lookups only when entry_number is present, thereby excluding entries with only two fields (prefix and reference).

The test makes sure that the entry with same existing reference is added when the organisation is different.


ssadhu-sl (approved):
lgtm

PR #220, opened 2024-07-19
Check categorical field types

JbannisterScottLogic:


digital_land/api.py
```diff
@@ -0,0 +1,74 @@
...
+        self.cache_dir = cache_dir
+
+    def download_dataset(self, dataset):
+        entities = []
+        count = 0
+        url = f"{self.url}/entity.json?dataset={dataset}&limit=500"
+
+        logging.info(f"Downloading dataset {dataset}")
+
+        while url:
```
eveleighoj, line 19: If you use the files cdn as mentioned above you don't need to do this batching. Also the name of url is confusing here as it starts out as a url and then you convert it into a response

digital_land/commands.py
```diff
@@ -233,8 +242,10 @@ def pipeline_run(
         ),
         HarmonisePhase(
             field_datatype_map=specification.get_field_datatype_map(),
+            specification=specification,
```
eveleighoj, line 245: Do we need the spec in here? (we may do but the field_datatype_map above is already passed in)

digital_land/commands.py
```diff
@@ -887,3 +898,40 @@ def organisation_check(**kwargs):
...
+    combined_matches = category_fields_query[
+        (category_fields_query["dataset"] == dataset)
+        | (category_fields_query["field"] == dataset)
+        | (category_fields_query["field-dataset"] == dataset)
+    ].drop_duplicates(subset=["dataset", "field", "field-dataset"])
+
+    if combined_matches.empty:
+        return
+
+    for _, row in combined_matches.iterrows():
```
eveleighoj, line 922: Does this still need to be in it looks like you've got this logic in the API call anyway

digital_land/phase/harmonise.py
```diff
@@ -170,3 +183,30 @@ def process(self, stream):
             block["row"] = o
 
             yield block
+
+    def get_category_fields(self):
```
eveleighoj, line 187: Should these be removed form here?

digital_land/phase/harmonise.py
```diff
@@ -93,6 +82,11 @@ def process(self, stream):
             o = {}
 
             for field in row:
+                if field in self.valid_category_values:
```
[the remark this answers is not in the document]
cjohns-scottlogic: self.valid_category_values is a dict of field : valid values. I think valid_category_values.keys() would be clearer.

ssadhu-sl (approved):
looks good

digital_land/api.py
```diff
@@ -0,0 +1,92 @@
...
+import requests
+import logging
+import typing
+
+DEFAULT_URL = "https://files.planning.data.gov.uk"
+
+
+class API:
+    def __init__(
+        self, specification, url: str = DEFAULT_URL, cache_dir: str = "var/cache"
```
ssadhu-sl, line 12: question: Do we always pass https://files.planning.data.gov.uk as the URL, what about when it is being executed on the dev/staging envt?

ssadhu-sl (commented):
Added one comment

ssadhu-sl (approved):

PR #221, opened 2024-07-22
Unique reference check

CarlosCoelhoSL:
Adds a check to ensure that the references in a resource are unique.
This is done by loading the fact csv generated by the pipeline into duckdb (after the pipeline is run), and checking whether there are duplicate references in the facts from that resource.
Duplicate references are only permitted under the condition the entry dates of the facts are different (as this would imply information has been updated in the resource)


digital_land/check.py
```diff
@@ -0,0 +1,37 @@
...
+        count_table = duckdb.sql(sql).df()
+        duplicate_references = count_table[count_table["count"] > 1]
+        for idx, row in duplicate_references.iterrows():
+            for entry_number in row["entry_numbers"].split(","):
+                issues.log_issue(
+                    "reference",
+                    "duplicate reference",
+                    row["value"],
+                    entry_number=int(entry_number),
+                    line_number=None,  # Can't get line number as we are outside pipeline
```
CarlosCoelhoSL, line 28: What do we pass in as line number here? Do we guess it based on the entry number? (ie line number = entry number + 1)

ssadhu-sl (dismissed):
Looks really good.

eveleighoj (approved):

PR #222, opened 2024-07-23
Update pipeline phases to accept endpoint hash/field

kena-SL:
Patch, Skip and Default phase updated to accept endpoint field.

Tests ensure records with endpoint are considered.


ssadhu-sl (approved):
lgtm

PR #223, opened 2024-07-26
Optimizing the duckDB check method

ssadhu-sl:
Reference ticket: https://trello.com/c/z4EIsWs5/3430-investigate-border-collection-failing

Border-collection action was failing due to a recent implementation of duplicate reference check. The change was functionally correct and working in most of the cases, but duckDb struggle with larger-than-memory data processing and was not able to handle complex csvs and was resulting in OOM issue. 

Changes Implemented:

- created temp table in duckDB
- Used dusk dataframe to fetch csv and filter data


CarlosCoelhoSL (approved):
lgtm

PR #224, opened 2024-07-26
Handle unhashable content.

cjohns-scottlogic:
Log errors generating hash from content, rather than crashing.


CarlosCoelhoSL (approved):
lgtm

PR #225, opened 2024-07-30
Adjust LookupPhase to include organisation while assigning lookups

kena-SL:
Ref: https://trello.com/c/klA9yNSh/1297-redo-adjust-pipeline-to-include-organisation-while-assigning-entities

For a new entity if a lookup with the same reference exists then pipeline doesnt seem to add a new lookup for the new entity as it considers it as already existing even when they belong to different organisation.

This PR updates the get_entity() of LookupPhase to check the entity number obtained using prefix and reference, it checks if an organisation is associated with the existing lookup if yes then it creates a new lookups for the incoming entity.

### DIFF
--- a/digital_land/phase/lookup.py
+++ b/digital_land/phase/lookup.py
@@ -57,11 +57,21 @@ class LookupPhase(Phase):
                 organisation=organisation,
                 reference=reference,
             )
+        )
+        if not entity:
             # TBD this needs to specifically not match unless the organisation and other columns
             # are empty in the lookups.csv probably isn't a change here.
             # or by the CURIE
-            or self.lookup(prefix=prefix, reference=reference)
-        )
+            entity = self.lookup(prefix=prefix, reference=reference)
+
+            if entity in self.lookups.values():
+                associated_keys = [
+                    key for key, value in self.lookups.items() if value == entity
+                ]
+                for key in associated_keys:
+                    parts = key.split(",")
+                    if len(parts) > 3 and parts[3]:
+                        entity = ""
 
         return entity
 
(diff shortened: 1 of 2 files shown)


digital_land/phase/lookup.py
```diff
@@ -57,11 +57,24 @@ def get_entity(self, block):
...
+            # When obtaining an entity number using only the prefix and reference, check if the
+            # lookup includes an associated organisation. If it does, do not use the entity number,
+            # as it is organisation specific.
+            if entity in self.lookups.values():
+                associated_keys = [
+                    key for key, value in self.lookups.items() if value == entity
+                ]
+                for key in associated_keys:
+                    parts = key.split(",")
+                    if len(parts) > 3 and parts[3]:
```
CarlosCoelhoSL, line 76: what does `key` look like here? I'm not totally sure, but even then I feel like there has to be a more specific way of checking whether there's an organisation present, e.g checking if 'local-authority' is in `parts` (or something like that, since I'm not 100% sure what keys looks like).

My concern with this way is how confident are we that `parts[3]` will always be the organisation, for instance if the columns are shuffled around in lookup.csv is it still that index?

CarlosCoelhoSL (commented):
lgtm just a couple of comments

digital_land/phase/lookup.py
kena-SL: Internally the lookups are formed using key() in [lookup.py](https://github.com/digital-land/digital-land-python/blob/main/digital_land/phase/lookup.py) which returns lookups this syntax - return ",".join([entry_number, prefix, reference, organisation])

It looks something like 
`{',ancient-woodland,abc0001,': '110000000', ',ancient-woodland,abc0001,local-authorityabc': '110000000'}`

Therefore the fourth element will always be an organisation.

We could check for if 'local-authority' is present in the org field, it could also have other values like development-corporation,national-park-authority or government-organisation. I have added a check for all of them

CarlosCoelhoSL (approved):
lgtm

PR #231, opened 2024-08-08
Adds new date pattern

CarlosCoelhoSL:
https://trello.com/c/7qz3aOif/3445-bug-dates-are-too-strict-and-dont-align-with-guidance

Some tests have had to be adjusted in accordance with the new 'priority' column being added to fact and fact-resource


ssadhu-sl (approved):
lgtm

PR #241, opened 2024-09-18
Added converted resource log.

cjohns-scottlogic:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add a record of the resource conversion process. Currently this is (mostly) handled by running ogr2ogr, but we do not currently record if this fails.

Some conversions are done in python, so it would be good to know if they fail (throw an exception) too.

In addition, the time to convert is recorded. This will help locate things which are taking a long time to convert.

- Ticket Link
- Related Issue #
- Closes #

images for UI changes._

## Added/updated tests?
_We encourage you to keep the code coverage percentage at 80% and above._

- [x] Yes
- [ ] No, and this is why: Test to be added

## [optional] Are there any post deployment tasks we need to perform?

Needs to be merged with matching makerules changes.

## [optional] Are there any dependencies on other PRs or Work?


digital_land/log.py
```diff
@@ -156,3 +156,43 @@ def add(self):
...
+class ConvertedResourceLog(Log):
+    fieldnames = [
+        "dataset",
+        "resource",
+        "command",
+        "env",
+        "gdal-version",
+        "return-code",
+        "stdout",
+        "stderr",
```
eveleighoj, line 170: Worth checking through the code to see if gdal is always used I think there are some where it's not.

Is it worth attempting to copy how downloading works a bit in terms of catching exceptions rather than stdout and stderr, they. can be included int he exception (we can create a special exception).

could be worth including the input format and maybe a result column so we know if it has. or hasn't been converted

digital_land/log.py
```diff
@@ -156,3 +156,43 @@ def add(self):
...
+                "env": self.env,
+                "gdal-version": self.gdal_version,
+                "return-code": self.return_code,
+                "stdout": self.stdout,
+                "stderr": self.stderr,
+            }
+        )
+
+    def save(self, *args, **kwargs):
+        self.add()
```
eveleighoj, line 197: Why is a blank row being added in the save bit?

digital_land/phase/convert.py
```diff
@@ -82,38 +82,50 @@ def read_excel(path):
...
+    rc, outs, errs = execute(command, env=env)
+
+    if log:
+        log.command = " ".join(command)
+        log.env = str(env)
+        log.gdal_version = str(gdal_version)
+        log.return_code = rc
+        log.stdout = outs
+        log.errs = errs
+
```
eveleighoj, line 128: Should log. be being added here

eveleighoj (approved):

PR #253, opened 2024-10-15
Implement checks in Entity Number assigning process

kena-SL:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR addresses the scenario where all entity numbers defined for a dataset have been used up.

**Current Behavior:**
When the code reaches the last available entity number, it loops back to the beginning and starts assigning numbers again without verifying whether the entity number already exists. This can result in merged entities, which is not correct.

**Updated Behavior:**
Before assigning an entity number, the code now checks if the number already exists. If it does, the number is skipped(this is for when new entities are added, we can always have duplicate entity number if we want). 
When all entity numbers are used, a message is printed to indicate that no more entity numbers are available for use.

- Ticket Link 
- Related Issue # 
- Closes #

## Added/updated tests?
_We encourage you to keep the code coverage percentage at 80% and above._

- [x] Yes
      have not been included_

### DIFF
--- a/digital_land/pipeline.py
+++ b/digital_land/pipeline.py
@@ -512,7 +512,7 @@ class Lookups:
 
                         if generated_entity == get_entity:
                             print(
-                                "There are no more numbers available within this dataset."
+                                "There are no more entity numbers available within this dataset."
                             )
                             break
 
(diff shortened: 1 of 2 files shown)


averheecke-tpx (approved):
lgtm

PR #299, opened 2024-12-04
Revert "Removed maximum size constraint when loading temp_table"

cjohns-scottlogic:
Reverts digital-land/digital-land-python#296


alexglasertpx (approved):
Reverting parquet changes

PR #300, opened 2024-12-04
Revert "Added incremental increase to max_line_size"

cjohns-scottlogic:
Reverts digital-land/digital-land-python#295


alexglasertpx (approved):
Reverting parquet changes

PR #303, opened 2024-12-04
Revert "Use parquet to speed up dataset building."

cjohns-scottlogic:
Reverts digital-land/digital-land-python#290


alexglasertpx (approved):
Reverting parquet changes

PR #310, opened 2025-01-02
Add data step 3

### LINKED ISSUE #309: Add Data Command - Add Lookups & Endpoints (#3)
**Overview**
- See full overview in this Mural: https://app.mural.co/t/mhclg2837/m/mhclg2837/1715695232673/731cd7c340728327108245664ddca4e6a27b95b7
- We want to build on the `add-data` command we have made in digital-land/digital-land-python#284 and digital-land/digital-land-python#308 as well as continue building acceptance tests for this command to ensure that future updates to the pipeline do not break this command.

**Tech Approach**
- For all these tickets, you should try and pair with one of the DM developers so that they can understand the code, advise on their working practices and (hopefully) take on some of the ongoing maintenance for it after implementation.
- Following previous ticket, we add a breakpoint to the command that asks the user whether they want to continue or not. At this point, they may want to select 'no' (in order to change config etc). At this point, the command will abort and delete any logs created.
- If they select 'yes', we should assign the entities and reprocess the resource.
- Check that the resource has been reprocessed correctly. If not (e.g. there are still unassigned entities), error out and print this to the console.
- If all goes well, print to the console:
    - Summary of new look-ups added (in lookup.csv)
    - min and max of new entity numbers
    - Count of new lookups
- Add a breakpoint that asks user if they want to save the new endpoint?
- If yes, save everything to the file
- If no, stop and don't save

**Acceptance Criteria**
- Digital Land command `add-data` can now add look-ups and endpoints for the transformed resource
- Acceptance tests that show if successful then resource is downloaded. If unsuccessful, command is exited and update to console log on why.
- Unit and/or integration tests added for any new (or altered) pieces of code.
- Testing by DM before going live.

CarlosCoelhoSL:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Step 3 for the add data command. In this step:
- the assign entities command is run
- the resource is processed again with assigned entities
- if there are remaining unassigned entities an exception is raised
- otherwise the user can save changes

- Related Issue #309 
- Closes #309

## Added/updated tests?

- [X] Yes
      have not been included_

### DIFF
--- a/digital_land/commands.py
+++ b/digital_land/commands.py
@@ -2,6 +2,7 @@ from collections import OrderedDict
 import csv
 import itertools
 import os
+import shutil
 import sys
 import json
 import logging
@@ -9,7 +10,7 @@ from packaging.version import Version
 import pandas as pd
 from pathlib import Path
 from datetime import datetime
-
+from distutils.dir_util import copy_tree
 import geojson
 from requests import HTTPError
 import shapely
@@ -712,7 +713,7 @@ def add_data(
     pipeline_dir,
     specification_dir,
     organisation_path,
-    cache_dir=None,
+    cache_dir=Path("var/cache/add_data/"),
 ):
     # Potentially track a list of files to clean up at the end of session? e.g log file
 
@@ -766,14 +767,14 @@ def add_data(
     operational_issue_dir.mkdir(parents=True, exist_ok=True)
 
     collection.load_log_items()
-    for pipeline in endpoint_resource_info["pipelines"]:
+    for dataset in endpoint_resource_info["pipelines"]:
         print("======================================================================")
         print("Run pipeline")
         print("======================================================================")
         try:
             pipeline_run(
-                pipeline,
-                Pipeline(pipeline_dir, pipeline),
+                dataset,
+                Pipeline(pipeline_dir, dataset),
                 Specification(specification_dir),
                 endpoint_resource_info["resource_path"],
                 output_path=output_path,
@@ -810,7 +811,7 @@ def add_data(
         print(f"Transformed resource path: {output_path}")
 
         column_field_summary = get_column_field_summary(
-            pipeline,
+            dataset,
             endpoint_resource_info,
             column_field_dir,
             converted_dir,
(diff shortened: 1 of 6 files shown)


digital_land/commands.py
```diff
@@ -712,7 +714,7 @@ def add_data(
     pipeline_dir,
     specification_dir,
     organisation_path,
-    cache_dir=None,
+    cache_dir=Path("var/cache/add_data/"),
```
eveleighoj, line 717: This could be set using an if clause if it's easier then changing how it's used.

Also should it create it's own directory. in the cache directory. I.e. I think we're moving towards a system that. expects a cache directory to be supplied and everything is using the cache directory to store stuff. So should the cache directory be "var/cache" and then the command automatically creates a folder in that directory to use.

So the code would then go add_data_cache = cache_dir / "add_data"

otherwise you'll have to pass a different cache directory to everyone of our commands which I reckon will get confusing.

digital_land/commands.py
```diff
@@ -766,14 +762,14 @@ def add_data(
     operational_issue_dir.mkdir(parents=True, exist_ok=True)
 
     collection.load_log_items()
-    for pipeline in endpoint_resource_info["pipelines"]:
+    for dataset in endpoint_resource_info["pipelines"]:
```
eveleighoj, line 765: endpoint_resource_info makes me nervous, it feels like something that should be easier. to get or better named. Don't worry about it for now but in future work it might be worth thinking about how we can make it easy to access this information.

digital_land/utils/add_data_utils.py
```diff
@@ -48,6 +48,16 @@ def is_date_valid(date, date_type):
     return True, ""
 
 
+def proceed(message):
```
eveleighoj, line 51: I'm not crazy about this name. maybe get_user_response. is better butt if you think proceed is easier then. leave it as that.

tests/acceptance/test_add_data.py
```diff
@@ -135,6 +135,21 @@ def mock_request_get(mocker):
     )
 
 
+@pytest.fixture
```
eveleighoj, line 138: In general you could look at adding some more tests

Unit:
* are all the util functions that. are small being unit tested?

Integration
* your using a lot of functions and methods where you could add integration tests at a more granular level (this is easier said than done)

CarlosCoelhoSL: A lot of of the functions (the get_issue/entity/column_field/_summary functions) already have integration tests from the previous PRs which cover the intended usage of the function. They don't have unit tests because they access the file system

I can definitely add some unit tests for the functions I've added that don't access the file system

eveleighoj (approved):

PR #369, opened 2025-03-18
Added function for update_dataset command (for incremental loading)

alexglasertpx:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

As part of incremental loading have added 'update_dataset' to the commands. Tests will come later once we start editing 'collection-pipeline-makerules'

- Ticket [Link](https://github.com/orgs/digital-land/projects/9/views/3?pane=issue&itemId=100509201&issue=digital-land%7Cdigital-land-python%7C361)

### DIFF
--- a/digital_land/makerules.py
+++ b/digital_land/makerules.py
@@ -56,7 +56,9 @@ def get_processing_option(
 
     # New resources downloaded
     if "resource" in diffs:
-        return ProcessingOption.PROCESS_ALL  # To be changed to partial in the future
+        return (
+            ProcessingOption.PROCESS_PARTIAL
+        )  # To be changed to partial in the future
 
     if not diffs:
         return ProcessingOption.PROCESS_NONE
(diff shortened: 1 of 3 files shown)


digital_land/makerules.py
```diff
@@ -56,7 +56,9 @@ def get_processing_option(
 
     # New resources downloaded
     if "resource" in diffs:
-        return ProcessingOption.PROCESS_ALL  # To be changed to partial in the future
+        return (
+            ProcessingOption.PROCESS_PARTIAL
+        )  # To be changed to partial in the future
```
CarlosCoelhoSL, line 61: With the way the rest of the pipeline_makerules() function works if this is returning PROCESS_PARTIAL it won't be printing out any `run-pipeline` statements because of this code:
```python
if process == ProcessingOption.PROCESS_ALL:
                    call_pipeline = (
                        "\t$(call run-pipeline,"
                        + " --endpoints '%s'" % endpoints
                        + " --organisations '%s'" % organisations
                        + " --entry-date '%s'" % entry_date
                    )
                    # we will include the resource arguement if the old resource
                    # is  different so it's processed as the old_resource
                    if resource != old_resource:
                        call_pipeline = call_pipeline + f" --resource '{old_resource}'"
                    call_pipeline = call_pipeline + " )"
                    print(call_pipeline)
```
We could remove the if statement from here I think and always print it, because to reach here the processing option is either PROCESS_ALL or PROCESS_PARTIAL, and we'll just filter the dataset_resource map for only the new resources when we have PROCESS_PARTIAL.

Also since get_processing_option() is now returning PROCESS_PARTIAL, which it never did before, we will need a new test that the pipeline_makerules() function can handle that (which would fail if it existed currently)

CarlosCoelhoSL (commented):
dataset-update command looks fine but the PROCESS_PARTIAL changes would cause pipeline.mk to not have run-pipeline statements printed currently if we only had new resources

digital_land/makerules.py
alexglasertpx: Have removed the if statememt

CarlosCoelhoSL (approved):
lgtm

PR #374, opened 2025-03-24
Update convert phase to handle windows-1252 encoded URLs

kena-SL:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Update convert phase to handle Windows-1252 encoded URLs.
Currently, add-data process expects the files to be UTF-8 encoded. When adding a file with Windows encoding it raise the below error

image

This PR updates the convert phase to ensure that files encoded in Windows-1252 are converted to UTF-8, so that they are processed correctly when adding data.

- Ticket Link https://mhclgdigital.atlassian.net/jira/software/projects/DATA/boards/229?selectedIssue=DATA-1327
- Related Issue #
- Closes #

images for UI changes._

## Added/updated tests?

- [x] Yes
      have not been included_

### DIFF
--- a/digital_land/phase/convert.py
+++ b/digital_land/phase/convert.py
@@ -15,6 +15,7 @@ from .load import Stream
 from .phase import Phase
 from ..utils.gdal_utils import get_gdal_version
 from digital_land.log import ConvertedResourceLog
+from requests import HTTPError
 
 
 class ConversionError(Exception):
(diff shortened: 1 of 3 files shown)


eveleighoj:
I think I might be a bit confused what this is aiming to fix. The file is encrypted with windows-1252. The convert phase is currently handling this fine (This might be where I'm confused) and is able to read it by examining the encoding first. 

Unless we want to change the behaviour (i.e. if it's not UTF-8 then created a converted csv) then it should be fine.

Do we instead want to just edit the add-data command to allow it to read differently encoded files? If the request was on the pipeline to do this I think your approach is okay but given we don't want the behaviour of the pipeline to change it might be best to improve how we read in the file in the add_data command.

There are two options:
- copy the aspects of the file reader code from convert phase that are capable of identifying the encoding and reading the file to the. add data code.  This is probably easier 
- refactor the conversion phase to separate out the file reader capability into a function. This function could be moved into the utils module. The function can then be used in the add_data command so that it matches what's ran in the conversion phase and any updates will be shared by both. this is possibly better but depends how hard it is to refactor the functionality out.

Let me know if I've misunderstood but unless we want to change the behaviour of the pipeline we probably shouldn't

PR #395, opened 2025-05-06
Feat/transport memory usage

### LINKED ISSUE #387: Transport-access-node collection failing.
Transport access node has failed for the last two days. Why is this happening and what can be done to resolve this?

alexglasertpx:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Edits to reduce the memory usage when building fact and fact resource parquet tables.

Closes #387 

## Added/updated tests?
- [X] Yes
      have not been included_


digital_land/package/dataset_parquet.py
```diff
@@ -136,23 +253,166 @@ def load_facts(self, transformed_parquet_dir):
...
+                        PARTITION BY fact ORDER BY priority, entry_date DESC, entry_number DESC
+                    ) = 1
+                """
+                self.conn.execute(
+                    f"""
+                    COPY (
+                        {query}
+                    ) TO '{str(output_path)}' (FORMAT PARQUET);
+                """
+                )
```
CarlosCoelhoSL, line 317: I can't help but feel there's a neater way of handling edge cases where there's only 1 batch/bucket (ie we're only running one query). It just feels like having 3 queries for 2 strategies suggests something might need to be refactored

I'm not sure where it's better to incorporate this edge case into the `self.strategy=="direct"` block or you could refactor the following else (multiple buckets and batched files) to handle only one batch/bucket.

The comment above implies in this edge case it it should function the same as `simple` strategy so it might be better to find a way to make sure they are using the same query

digital_land/package/dataset_parquet.py
```diff
@@ -177,6 +445,148 @@ def load_fact_resource(self, transformed_parquet_dir):
         """
         )
 
+    # def load_details_into_temp_parquet(self, transformed_parquet_dir):
```
CarlosCoelhoSL, line 448: why is this function and its test commented out?

CarlosCoelhoSL (commented):
looks good, a few parts i think need to be refactored.
also there are quite a few commented out pieces of code (not just the function) that could be tidied

digital_land/package/dataset_parquet.py
```diff
@@ -64,6 +69,13 @@ def __init__(self, dataset, path, duckdb_path=None, **kwargs):
         self.entity_path = (
             self.path / "entity" / f"dataset={self.dataset}" / "entity.parquet"
         )
+        if transformed_parquet_dir is None:
+            self.strategy = "direct"
+        else:
+            self.parquet_dir_details = self.analyze_parquet_dir(
+                transformed_parquet_dir=transformed_parquet_dir
+            )
+            self.strategy = self.choose_strategy(self.parquet_dir_details)
```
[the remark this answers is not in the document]
alexglasertpx: `self.parquet_dir_details` are also called in the `load_facts` function.

digital_land/package/dataset_parquet.py
alexglasertpx: Originally there were 3 strategies but this was reduced to 2. Perhaps we can think of this as a two-step process:

1. is the data batched or not? (if not then the strategy is 'direct')
2. if it is batched then do we need to bucket it to run the windowed SQL function?

When we have one bucket or batched file then the query is slightly different (reading in the batch files rather than the original parquet files). Some of the logic used (like calculating `n_buckets`) could be earlier but if the strategy = 'direct' then it is redundant.

eveleighoj (approved):

PR #532, opened 2026-04-28
add_count func refactored and unittests

salabi:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This change moves entity_count calculation from the dataset packaging step into the transform flow, so the count is calculated per resource as soon as the transformed file is produced. The result is written into the dataset-resource log and then loaded into the package, removing the need for the later SQLite add_counts() update.

As part of this, the old package.add_counts() calls were removed from dataset create/update, and the transform pipeline now computes entity_count directly from the transformed resource by counting distinct non-empty entity values. Tests were updated to reflect the new flow and to ensure the count is preserved when the dataset-resource log is loaded 

- Ticket Link https://github.com/orgs/digital-land/projects/15/views/1?pane=issue&itemId=169411811&issue=digital-land%7Cconfig%7C2421
- Related Issue #
- Closes #

images for UI changes._

## Added/updated tests?

- [ ] Yes
      have not been included_


digital_land/pipeline/main.py
```diff
@@ -56,6 +56,18 @@ class PipelineStatus(Enum):
     FAILED = 5
 
 
+def count_distinct_entities(csv_path):
```
eveleighoj, line 60: This approach is going to be incredibly inefficient. We're already iterating through each row for the transformation step. there are two main options:
- calculate this list in the entity lookup phase or somewhere and then at the end do this calculation. Although this will use up more memory when running
- use duckdb or polars to calculate this, both can offer efficient calculations on larger datasets

digital_land/package/dataset.py
```diff
@@ -182,29 +182,6 @@ def load_entities(self):
         self.commit()
         self.disconnect()
 
-    def add_counts(self):
```
eveleighoj, line 185: don't need to delete this yet, can leave as part of the codebase.

eveleighoj (changes requested):
we need to be confident that the method of calculating the unique count is efficient

digital_land/pipeline/main.py
salabi: resolved. Using polars to do the counting now.

digital_land/package/dataset.py
salabi: updated.

eveleighoj (approved):

PR #539, opened 2026-05-07
fix: remove split_curie from EntityReferencePhase

### LINKED ISSUE #729: Remove prefix check for CURIE in references
### Summary Description
When adding an endpoint or assigning entities to an endpoint, if the reference has a :(colon) in the reference then the it is process as prefix:reference 
For e.g. if the reference value is `DA:20/21=S106:001:20/0284/FUL` then the assign-entities process considers it has prefix:reference

We have created an issue type to monitor these references so we can analyse the extent of the problem and then plan our next steps. This can be found at: https://datasette.planning.data.gov.uk/digital-land/issue?issue_type=reference+value+contains+reference_prefix&_sort=dataset

### Technical Approach

* The entity reference phase for some reason attempts to split the curie from the reference column. This wouldn't be a problem if the reference wasn't able to contain ':' itself. because it is able to contain ':' we are seeing that some references are being split when they shouldn't
* Are there are examples in our code base where this is being used? might need to add some logging to check as we might not know what we're breaking if we turn it off. **This has been implemented in the parent ticket!** Check a the set of collections that are shown in the issue type including:
  * brownfield-land
  * Development agreement transaction/contribution

* once we have checked these collections - if we see no examples of where this is used we should look at adjusting the phase (not just putting a hot fix on it) we should only attempt to use the reference prefix if sef.prefix isn't supplied. **This may be required to be raised as a new ticket for development**
* if there are examples of this we need to review what to do. things to consider:
  * are the prefixes supplied correct? are we actually seeing errors where an incorrect prefix is being used in our data?
  * does this alter the reference as well so do we need to change/add lookups where the wrong prefix is being used

tombrooks248:
## Description

Removing function `split_curie` as part of the `EntityReferencePhase` 

Removing issue logging for reference prefix

## Why

`split_curie` because it is no longer (nor ever has been) been required as data providers have never needed to provide references as a Curie.

Issue logging no longer required now we have identified the issues and that there is no need for this Curie handling code.

- [Ticket Link](https://github.com/orgs/digital-land/projects/15/views/1pane=issue&itemId=113801168&issue=digital-land%7Cconfig%7C729)
- Related Issue #
- Closes #

After discussing with Ben we are sure that there are in fact no data providers who are sending us curies and, given that the splitting is catching several colon-containing-references by mistake, we have deemed removing the function to be safe.

## Added/updated tests?

- [x] Yes
      have not been included_

### DIFF
--- a/tests/unit/phase/test_reference.py
+++ b/tests/unit/phase/test_reference.py
@@ -5,7 +5,6 @@ import pytest
 from digital_land.specification import Specification
 from digital_land.phase.reference import EntityReferencePhase, FactReferencePhase
 from digital_land.phase.reference import split_curie
-from digital_land.log import IssueLog
 
 
 def test_split_curie():
@@ -22,14 +21,10 @@ def test_entity_reference():
     assert ("foo", "Q1234") == phase.process_row(
         {"prefix": "foo", "reference": "Q1234"}
     )
-    assert ("wikidata", "Q1234") == phase.process_row({"reference": "wikidata:Q1234"})
     assert ("tree", "NSP22: Not A CURIE") == phase.process_row(
         {"reference": "NSP22: Not A CURIE"}
     )
     assert ("tree", "Not A CURIE:") == phase.process_row({"reference": "Not A CURIE:"})
-    assert ("foo", "Q1234") == phase.process_row(
-        {"prefix": "foo", "reference": "wikidata:Q1234"}
-    )
 
 
 def test_fact_reference():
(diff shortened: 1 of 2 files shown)


tests/unit/phase/test_reference.py
```diff
@@ -22,14 +21,10 @@ def test_entity_reference():
     assert ("foo", "Q1234") == phase.process_row(
         {"prefix": "foo", "reference": "Q1234"}
     )
-    assert ("wikidata", "Q1234") == phase.process_row({"reference": "wikidata:Q1234"})
```
gibahjoe, line 25: Instead of removing these tests, could we invert them so they explicitly check that the value is *not* this instead?

gibahjoe (commented):
It makes future changes clearer: if someone later introduces the split again, the test will fail immediately.

tests/unit/phase/test_reference.py
tombrooks248: Sure good idea - made that change.

eveleighoj (approved):

PR #547, opened 2026-05-13
feat: add new TaskPipeline Class

tombrooks248:
## Description

Add a new TaskPipeline Class for use by the check service as well as other processes in future.

## Why

Currently the check service exists to figure out what "Tasks" LPAs have to do to fix the various data errors which exist in their data that they are using the check service to check. The front end kicks off an Async process which generates a lot of logs/issues about any data errors. Once done the front-end code runs some code to consolidate these logs/issues so that it can be presented nicely to the user. e.g. a one line summary of 400 identical errors to be fixed rather than 400 rows of something.

We are attempting to recreate this consolidation code in this repository so that it can run as part of the Async task rather than in the front-end code, because we think this will make the process faster and also enable the code to be reused in other processes.

- [Ticket Link](https://github.com/orgs/digital-land/projects/15/views/1?pane=issue&itemId=179396849&issue=digital-land%7Cdigital-land-python%7C526)
- Related Issue #
- Closes #

images for UI changes._

## Added/updated tests?

- [x] Yes
      have not been included_


digital_land/pipeline/task.py
```diff
@@ -0,0 +1,293 @@
...
+    "dataset",
+    "organisation",
+    "endpoint",
+    "resource",
+    "details",
+    "severity",
+    "responsibility",
+    "task-source",
+    "entry-date",
+    "task-id",
```
eveleighoj: I am going to have to push back on task-id. We should use 'reference'

tests/unit/test_task_pipeline.py
```diff
@@ -0,0 +1,183 @@
+import csv
```
eveleighoj: So it's worth a read through https://digital-land.github.io/technical-documentation/development/testing-guidance/

tests for this class and methods should be in tests/unit/pipeline/test_task.py

then all unit tests for that task are found in the same place

then the structure for the run method tests should be:

```
class TestTaskPipeline:

    def test_run_<behavior>(sef):
``` 

keeps all tests for that method grouped together under a single test class. if methods are duplicated across classes in the same python file then keeps it obvious which one it's testing. All tests for that method are easily found

digital_land/pipeline/task.py
```diff
@@ -0,0 +1,293 @@
...
+                ).encode()
+            ).hexdigest()[:16],
+            return_dtype=pl.Utf8,
+        )
+        .alias("task-id")
+    )
+    return result
+
+
+def _tasks_from_issues(
```
eveleighoj: Could make this more of a transform function and instead the pipeline run method can do the scan_csv part. Then makes it easier to add some specific unit tests for this function where you just feed in polars data frames. The transformation logic will become separated from the reading logic

tests/unit/test_task_pipeline.py
tombrooks248: Sure done this, but do take a look, and check its right.

digital_land/pipeline/task.py
tombrooks248: Sure had a go at this, do check it out.

digital_land/pipeline/task.py
tombrooks248: Sure, made that change.

pooleycodes (approved):

pooleycodes (approved):

PR #548, opened 2026-05-14
Implement the NULL resource_end_date bug fix

sianteesdale:
This PR was previously #528 and #531. These PRs were reverted due to believing they were causing hour long delays in Airflow every night. Subsequent investigations by @tombrooks248 and @salabi (https://github.com/orgs/digital-land/projects/15?pane=issue&itemId=161656529&issue=digital-land%7Cconfig%7C2286) found that the delays were caused by the way the code was retrieving data from Datasette.

I am now re-testing if their edits to the code will allow my NULL bug fix to be implemented without any additional delays.
We're still keen to fix the NULL date, given it is causing 1000s of entities to be recorded as deleted erroneously.

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR addresses a bug found within the deleted entities expectation.

1. resource_end_date filter
Active resources are fetched from reporting_historic_endpoints using a filter on resource_end_date. Some active resources store NULL in this field rather than an empty string, so they were being excluded. The filter was updated to (resource_end_date = '' OR resource_end_date IS NULL).

- Ticket Link: https://github.com/orgs/digital-land/projects/15/views/1?pane=issue&itemId=161656529&issue=digital-land%7Cconfig%7C2286

Run the integration tests for count_deleted_entities:

`pytest tests/integration/expectations/operations/test_dataset.py -v -k "deleted"`

This covers:
- The baseline case (one entity with no active facts is reported deleted)
- All entities active — check passes with 0 deleted
- Entities with empty reference strings — regression test for the GROUP BY collapse bug
- Organisation isolation — facts from a second org's resource don't affect the queried org's count

## Added/updated tests?
- [X] Yes
      have not been included_

### DIFF
--- a/digital_land/expectations/operations/dataset.py
+++ b/digital_land/expectations/operations/dataset.py
@@ -126,6 +126,7 @@ def count_lpa_boundary(
 
     return result, message, details
 
+
 # Previously named count_deleted_entities() function
 def fetch_active_resources_for_dataset(dataset_name):
     params = urllib.parse.urlencode(
--- a/digital_land/expectations/operations/dataset.py
+++ b/digital_land/expectations/operations/dataset.py
@@ -133,7 +133,7 @@ def fetch_active_resources_for_dataset(dataset_name):
             "sql": f"""select o.entity as organisation_entity, rhe.resource
                         from reporting_historic_endpoints rhe
                         join organisation o on rhe.organisation=o.organisation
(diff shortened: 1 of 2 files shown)


salabi (approved):
LGTM

PR #554, opened 2026-05-19
Add column field log to task pipeline

pooleycodes:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Extends `TaskPipeline` to detect and surface missing mandatory fields from a
column-field log as task rows.

A new `column_field_path` parameter (alongside `mandatory_fields`) can be passed
to `TaskPipeline.run()`. When provided, the pipeline reads the column-field CSV,
compares the `field` column against the supplied mandatory fields list, and emits
one `column-field` task row per missing field — using the same schema and SHA-256
`reference` hash as the existing `log` and `issue` sources.

Key behaviours:
- `mandatory_fields` entries can be a plain string (field must be present) or a
  list of strings (any one of the alternatives is sufficient)
- If `column_field_path` is given without `mandatory_fields`, a warning is logged
  and the check is skipped gracefully
- If the path does not exist, a warning is logged and it is skipped
- Fixes `_transform_issues_to_tasks` to guard against `None` organisation/endpoint
  (was passing `None` directly to `pl.lit`)
- `run()` now consistently returns `self.status` rather than the bare enum value

- Ticket Link https://github.com/orgs/digital-land/projects/15?pane=issue&itemId=179396849&issue=digital-land%7Cdigital-land-python%7C526

## QA Instructions

Pass a column-field CSV (columns: `dataset`, `resource`, `column`, `field`) and a
`mandatory_fields` list to `TaskPipeline().run()`. Rows for any fields absent from
the CSV will appear in the output with `task-source=column-field` and `details`
containing `{"field": "...", "issue_type": "missing-field"}`.

## Added/updated tests?

- [x] Yes — `TestColumnFieldTasks` in `tests/unit/pipeline/test_task.py`

### DIFF
--- a/digital_land/pipeline/task.py
+++ b/digital_land/pipeline/task.py
@@ -64,6 +64,8 @@ class TaskPipeline:
         endpoint: Optional[str] = None,
         log_path: Optional[Path] = None,
         issue_path: Optional[Path] = None,
+        column_field_path: Optional[Path] = None,
+        mandatory_fields: Optional[List[str]] = None,
         severity_filter: Optional[List[str]] = None,
         responsibility_filter: Optional[List[str]] = None,
         entry_date: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


gibahjoe (approved):
LGTM

PR #572, opened 2026-06-30
feat: add new lookup-rule for  entity numbers

tombrooks248:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds a **lookup rule** mechanism for assigning entity numbers. Instead of an exact `lookup.csv` row per reference, a collection can provide an optional `lookup-rule.csv` defining a rule: apply an `offset` to an incoming integer reference, and if the result falls within a dataset's entity range (`entity-minimum`/`entity-maximum`, inclusive), assign that number as the entity.

## Why

Some datasets - e.g. title-boundaries - would require managing an unmanageable number of individual entity records if we relied on exact reference-to-entity lookups. A range-based rule lets a whole dataset's entities be assigned from a single, compact rule instead of millions of lookup rows, which is the prerequisite for adding title boundaries.

- [Ticket Link](https://github.com/digital-land/config/issues/2748)
- Related Issue #
- Closes #

True testing for this will take place once I begin running title-boundary in DEV, but until then this is going to lie dormant for a few days until that is ready.

## Testing this PR is safe

This change is **inert for existing pipelines** — it cannot alter the output of any current collection, because none of them have `lookup_rule.csv`s in the config repo, in fact currently no collections have lookup_rule.csv in the config repo and this code change will only do anything when I add one to title-boundry.  

Ran a pipeline locally with the changes and it all ran fine.

Plus all the existing tests pass.

## Added/updated tests?

- [x] Yes
      have not been included_

### DIFF
--- a/digital_land/phase/lookup.py
+++ b/digital_land/phase/lookup.py
@@ -33,6 +33,7 @@ class LookupPhase(Phase):
         issue_log=None,
         operational_issue_log=None,
         entity_range=[],
+        lookup_rules=None,
     ):
         self.lookups = lookups
         self.redirect_lookups = redirect_lookups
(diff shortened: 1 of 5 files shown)


gibahjoe (approved):
LGTM

Issue #573, opened 2026-07-07
Pipeline support for multi-organisation resources


Ben-Hodgkiss:
Currently the platform treats a resource as a globally unique artefact identified solely by its
content hash. When two organisations submit the same URL they produce the same hash, and the collection merges them into a single resource record with both organisation codes. The pipeline has no way to process that resource twice — once per organisation — so entity lookup silently fails for both.

To handle this correctly at the platform level the pipeline would need to be extended to support multi-organisation resources. When a resource is associated with more than one organisation, the pipeline would process it once per organisation, setting each organisation as the default in turn and writing out a separate set of transformed rows per organisation. This would allow the existing lookup.csv entries to resolve correctly without any change to the submission.

This is a non-trivial change given that the resource uniqueness model is fundamental to how the collection, transformation, and packaging steps are currently designed

tombrooks248:
THis is a good example - it not working https://provide.planning.data.gov.uk/organisations/local-authority:WYC

There should not be zero in local plan or plan timetable.  for the field - Number of records supplied

We have 2 endpoints - which are providing the same data - but the pipeline is not handling that correctly.

and as a consequence some entities are not being created which should be. resulting in missing data e.g. on the LPA dashboard above.

We think this might be part of the lookup phase - which is failing to find/connect the resources with all of the endpoints that generate it.

We have entities which are defined in the lookup.csv - but we are not creating entities for.

tombrooks248:
Clarification while investigating. 

The issues here not just that there are multiple endpoints serving the same resources, its that those multiple endpoints are from different **organisations**. Making new rows to connect all the orgs to a given resource should be easy - the main work of this ticket is figuring how what other bits of our pipelines make the assumption that each resource only comes from one organisation, and if we change that what needs to be changed to ensure it still all runs smoothly.

tombrooks248:
ok so I now have a PR to fix this, and tested it in Staging and it seemed like it started creating entities that we needed creating. I have sent to @eveleighoj for review, because its a fairly core bit of the pipelines I am changing so I think should get a second opinion.

tombrooks248:
Ok so Owen gave me a bunch of feedback, and while this is mostly doing what we want it to, but it doesn't always and its a bit more complicated that it needs to be. So I am moving it back to in-flight to have a second attempt at building something which achieves the same end but fits the architecture better.

tombrooks248:
Owen Approved - code merged. moving to done.

[...]

PR #574, opened 2026-07-07
feat: handle multiple orgaisations per resource

tombrooks248:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Changing the way the pipeline generates entities from a resource in a situation where multiple endpoints **from different organisations** provide identical data and therefore produce the same resource.

Before - No entities are produced from resources with multiple orgs.
After - now generate entities per organisation per resource. 

## Why 

Because we are currently missing entities in our data and this is causing a bunch of problems particularly for local-plan data.

## How the bug works

A resource is identified only by the hash of its contents. When two organisations publish the same URL, they produce identical bytes, so the collection merges them into a **single resource that carries both organisation codes**.

To generate entities, the pipeline stamps a default organisation onto every row of a resource (the data files themselves don't say which authority they belong to). But it only does this when there is exactly one organisation:

```python
if len(organisations) == 1:
    default_values["organisation"] = organisations[0]
```

## Worked example — South & Vale Joint Local Plan 2045

South Oxfordshire and Vale of White Horse jointly publish one local plan timetable at a single URL:

`https://www.southandvale.gov.uk/app/uploads/2026/06/Local-Plan-Timetable-CSV.csv`

In `source.csv` that one endpoint is attributed to both authorities:

| organisation | dataset |
|---|---|
| `local-authority:SOX` | plan-timetable |
| `local-authority:VAL` | plan-timetable |

`lookup.csv` already defines separate entities for each authority against the same references, for example:

| dataset | reference | SOX entity | VAL entity |
|---|---|---|---|
| plan-timetable | `Joint-Local-Plan-2045-scoping-consultation-start` | 5110200 | 5110188 |
| local-plan | `south-and-vale-joint-local-plan-2045` | 4221014 | 4221015 |

**Before:** the resource carries two organisations, the guard is skipped, and neither authority's entities are created — both show zero records supplied.

**After:** the resource is processed once per organisation (each set as the default in turn), both lookups resolve, and both authorities get their entities.

- [Ticket Link](https://github.com/digital-land/digital-land-python/issues/573)
- Related Issue #
- Closes #

### Downstream impacts of this change.

Processing a resource once per organisation means a single resource now produces rows for more than one organisation. I looked at every step that runs after transform to check nothing relies on the old "one organisation per resource" assumption.

#### Handled by this change

- **Duplicate-reference check.** This flags a reference appearing more than once in a resource. With two organisations, the same reference legitimately appears for each (e.g. the same joint-plan milestone), which would be falsely flagged as "not unique". We now run this check **per organisation pass**, where references are genuinely expected to be unique, so no false positives are introduced.

- **Doubled logs.** The org-independent side-logs (column-field, converted-resource) would otherwise be written once per pass. We write the converted/harmonised artefacts on the first pass only and de-duplicate those log rows, so a multi-organisation resource produces the same single set of log rows as any other.

- **Issue → entity attribution.** Issues are logged before the entity is known and back-filled afterwards. We apply and reset that mapping per pass so each organisation's issues attach to that organisation's entities rather than the other's.

#### Not affected (checked, no change needed)

- **Dataset / entity assembly.** Entities are assembled by grouping on the entity number, not on the resource. Each organisation resolves to its own entity, so the two organisations' rows become two distinct entities — exactly as if they'd come from separate resources. The organisation is carried through as a normal field.

- **Priority.** Priority is decided per entity from that entity's authoritative organisation. Each organisation owns its own entity, so both resolve correctly.

- **Task generation.** Tasks are keyed by dataset/resource/field/issue-type/organisation and already expand a resource's organisations into one task per organisation. They take the organisation from the resource, not from anything this change touches, so they are unaffected.

- **Single / no-organisation resources.** The vast majority of resources have one organisation (or none). Those follow the same path as before — this change only alters behaviour when a resource has two or more organisations.

### Testing in Staging

I deployed this to staging and ran the `local-plan` collection through to the plan-timetable Postgres load. Before this change the South & Vale Joint Local Plan produced no entities; after it, querying the staging `entity` table returns all 12 plan-timetable milestones **twice** — once for South Oxfordshire and once for Vale of White Horse — with the exact entity numbers defined in `lookup.csv` (e.g. `Joint-Local-Plan-2045-scoping-consultation-start` → 5110200 for South Oxfordshire and 5110188 for Vale of White Horse). That confirms both authorities now get their own entities from the single shared resource, end to end through transform, dataset build and the Postgres loader.

### Why this is safer than it first looks

Changing how resources map to organisations sounds like it touches a core assumption, but the pipeline is already **entity-centric below the resource level** — facts, dataset assembly and tasks are all keyed on the entity, not the resource. So "process a resource once per organisation" fits the existing model rather than fighting it: each organisation resolves to its own entity, and from that point on everything treats the two organisations exactly as if they'd come from separate resources.

The load-bearing detail is that a fact's identity is a hash of `entity : field : value`, so two organisations producing identical rows still create distinct facts (different entities → different hashes). That's why nothing collapses or overwrites. I also confirmed the resource-level logs are set, not accumulated, so nothing double-counts.

The change is also naturally bounded: resources with one organisation (or none) follow exactly the same path as before — only resources with two or more organisations hit the new behaviour. Which is a relatively small subset of the total data.

## Added/updated tests?

- [x] Yes
      have not been included_

## Summary by CodeRabbit

* **New Features**
  * Multi-organisation resource processing now combines results from each organisation into one final output.
  * Output handling is more consistent across single-, zero-, and multi-organisation runs.

* **Bug Fixes**
  * Duplicate log entries are now reduced in the final output.
  * Organisation-specific issues no longer leak between processing passes, improving accuracy in multi-organisation runs.

### DIFF
--- a/digital_land/pipeline/main.py
+++ b/digital_land/pipeline/main.py
@@ -72,6 +72,46 @@ def count_distinct_entities(csv_path):
     )
 
 
+def _concat_transformed(input_paths, output_path):
+    """Concatenate per-organisation transformed CSVs into one resource file.
+
+    Writes the header once (from the first file) then the data rows of every
+    file, and removes the per-organisation inputs. Used for multi-organisation
+    resources, where the pipeline is run once per org and the outputs are
+    stitched into the single resource-keyed file the rest of the pipeline expects.
+    """
+    output_path = Path(output_path)
+    with open(output_path, "w", newline="") as out:
+        for idx, path in enumerate(input_paths):
+            with open(path, newline="") as f:
+                header = f.readline()
+                if idx == 0:
+                    out.write(header)
+                for line in f:
+                    out.write(line)
+    for path in input_paths:
+        if Path(path) != output_path:
+            os.remove(path)
+
+
+def _dedup_log_rows(log):
+    """Collapse duplicate rows in an org-independent per-resource side-log.
+
+    The column-field and converted-resource logs are appended once per
+    organisation pass, so a multi-organisation resource produces identical
+    duplicate rows. Deduplicate them, preserving order. Assumes flat rows
+    (string/scalar values).
+    """
+    seen = set()
+    deduped = []
+    for row in log.rows:
+        key = tuple(sorted(row.items()))
+        if key not in seen:
+            seen.add(key)
+            deduped.append(row)
+    log.rows = deduped
+
+
 def chain_phases(phases):
     def add(f, g):
         return lambda x: g.process(f(x))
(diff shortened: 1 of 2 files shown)


tombrooks248:
This solution involved looping over each resource, we decided to use a differnt architecture to solve the same problem in a different PR so this PR can now be closed.

PR #577, opened 2026-07-13
feat: handle multiple organisation per resource

tombrooks248:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Changing the way the pipeline generates entities from a resource in a situation where multiple endpoints **from different organisations** provide identical data and therefore produce the same resource.

Before - No entities are produced from resources with multiple orgs.
After - now generate entities per organisation per resource. 

## Why 

Because we are currently missing entities in our data and this is causing a bunch of problems particularly for local-plan data.

## How the bug works

A resource is identified only by the hash of its contents. When two organisations publish the same URL, they produce identical bytes, so the collection merges them into a **single resource that carries both organisation codes**.

To generate entities, the pipeline stamps a default organisation onto every row of a resource (most data files don't say which authority they belong to). But it only does this when there is exactly one organisation:

```python
if len(organisations) == 1:
    default_values["organisation"] = organisations[0]
```

## Worked example — South & Vale Joint Local Plan 2045

South Oxfordshire and Vale of White Horse jointly published one local plan timetable at a single URL:

`https://www.southandvale.gov.uk/app/uploads/2026/06/Local-Plan-Timetable-CSV.csv`

In `source.csv` that one endpoint was attributed to both authorities:

| organisation | dataset |
|---|---|
| `local-authority:SOX` | plan-timetable |
| `local-authority:VAL` | plan-timetable |

`lookup.csv` already defines separate entities for each authority against the same references, for example:

| dataset | reference | SOX entity | VAL entity |
|---|---|---|---|
| plan-timetable | `Joint-Local-Plan-2045-scoping-consultation-start` | 5110200 | 5110188 |

**Before:** the resource carries two organisations, the guard is skipped, and neither authority's entities are created — both show zero records supplied.

**After:** entity lookup resolves the reference once per organisation and emits a row per distinct entity, so both authorities get their entities.

- [Ticket Link](https://github.com/digital-land/digital-land-python/issues/573)
- Related Issue #
- Closes #

### Tested in DEV

DEV's collected `resource.csv` has **4 active multi-org resources** (organisations column containing `;`). I tested by looking at this one:

|-|-|
|---|---|
| Resource | `ae8f59326ef53ede176817221ff512c00d3bc860874e8b8add2fb84dbc06e1f5` |
| Organisations | `local-authority:ADU;local-authority:WOT` (Adur + Worthing) |
| Endpoint | `ea98ea4d156ee47f…` |
| Dataset | brownfield-land |

Adur and Worthing publish a single shared brownfield-land CSV. 

#### Result — the fix works

- **743 fact rows, 45 distinct entities, 0 blank-entity rows**
- `local-authority:ADU` — 23 entities
- `local-authority:WOT` — 22 entities
- all rows `priority=2` (authoritative)

**Before the change this file was empty** — no organisation meant no entity, and every row was pruned. Both councils showed zero records supplied. Both councils' data is now created from the single shared resource.

### Why this is safer than it first looks

Changing how resources map to organisations sounds like it touches a core assumption, but the pipeline is already **entity-centric below the resource level** — facts, dataset assembly and tasks are all keyed on the entity, not the resource. So "process a resource once per organisation" fits the existing model rather than fighting it: each organisation resolves to its own entity, and from that point on everything treats the two organisations exactly as if they'd come from separate resources.

The load-bearing detail is that a fact's identity is a hash of `entity : field : value`, so two organisations producing identical rows still create distinct facts (different entities → different hashes). That's why nothing collapses or overwrites. I also confirmed the resource-level logs are set, not accumulated, so nothing double-counts.

The change is also naturally bounded: resources with one organisation (or none) follow exactly the same path as before — only resources with two or more organisations hit the new behaviour. Which is a relatively small subset of the total data.

### Impact on issues generated
an unknown-entity issue is raised only when no organisation resolves an entity

## Added/updated tests?

- [x] Yes
      have not been included_

## Summary of changes

* **New Features**
  * Entity lookups can now fan out across multiple organisations, stamping the resolved organisation onto both the output row and block.
  * If an input row already specifies its organisation, lookups respect it without fan-out.

* **Bug Fixes**
  * Duplicate reference detection is now evaluated separately per entity, and duplicate reports include the correct entity context.
  * “Unknown entity” handling is more consistent, including missing-reference cases, with only one issue raised when no organisations resolve.

* **Documentation**
  * Updated pipeline documentation to clarify how multi-organisation lookup behaves.

* **Tests**
  * Added unit coverage for multi-organisation fan-out, deduplication, and unknown-match scenarios, plus an extra duplicate-reference case.

### DIFF
--- a/digital_land/phase/lookup.py
+++ b/digital_land/phase/lookup.py
@@ -84,14 +84,14 @@ class LookupPhase(Phase):
                     return ""
         return entity
 
-    def get_entity(self, block):
+    def get_entity(self, block, organisation=None):
         row = block["row"]
         prefix = row.get("prefix", "")
         reference = row.get("reference", "")
         # Eventually won't be needed but leave in for now
-        organisation = row.get("organisation", "").replace(
-            "local-authority-eng", "local-authority"
-        )
+        if organisation is None:
+            organisation = row.get("organisation", "")
+        organisation = organisation.replace("local-authority-eng", "local-authority")
         entry_number = block["entry-number"]
 
         entity = (
@@ -143,6 +143,31 @@ class LookupPhase(Phase):
                     return ""
         return entity
 
+    def _log_unknown_entity(self, reference, curie, line_number):
+        if not self.issues:
+            return
+        if not reference:
+            self.issues.log_issue(
+                "entity",
+                "unknown entity - missing reference",
+                curie,
+                line_number=line_number,
+            )
+        else:
+            self.issues.log_issue(
+                "entity",
+                "unknown entity",
+                curie,
+                line_number=line_number,
+            )
+            if self.operational_issues:
+                self.operational_issues.log_issue(
+                    "entity",
+                    "unknown entity",
+                    curie,
+                    line_number=line_number,
+                )
+
     def process(self, stream):
         for block in stream:
             row = block["row"]
@@ -189,13 +214,84 @@ class LookupPhase(Phase):
 class EntityLookupPhase(LookupPhase):
     entity_field = "entity"
 
+    def __init__(
+        self,
+        lookups={},
+        redirect_lookups={},
+        issue_log=None,
+        operational_issue_log=None,
+        entity_range=[],
+        lookup_rules=None,
+        organisations=None,
+    ):
+        super().__init__(
+            lookups=lookups,
+            redirect_lookups=redirect_lookups,
+            issue_log=issue_log,
+            operational_issue_log=operational_issue_log,
+            entity_range=entity_range,
+            lookup_rules=lookup_rules,
+        )
+        self.organisations = organisations or []
+
     def process(self, stream):
+        # A resource may be associated with more than one organisation (e.g. a
+        # joint local plan shared by two authorities). In that case the entity
+        # is resolved once per organisation and a row is emitted per distinct
+        # entity. With a single organisation (or none) each row resolves to at
+        # most one entity.
+        if len(self.organisations) > 1:
+            yield from self._process_multi_org(stream)
+            return
+
         for block in super().process(stream):
             if self.issues:
                 self.issues.record_entity_map(
                     block["entry-number"], block["row"]["entity"]
                 )
             yield block
+
+    def _process_multi_org(self, stream):
+        for block in stream:
+            row = block["row"]
+            prefix = row.get("prefix", "")
+            reference = row.get("reference", "")
+            curie = f"{prefix}:{reference}"
+            line_number = block["line-number"]
+
+            # Nothing to look up if there is no prefix, or an entity has
+            # already been assigned upstream.
+            if not prefix or row.get(self.entity_field, ""):
+                if self.issues:
+                    self.issues.record_entity_map(
+                        block["entry-number"], row.get(self.entity_field, "")
+                    )
+                yield block
+                continue
+
+            emitted_entities = set()
+            for organisation in self.organisations:
+                entity = self.get_entity(block, organisation=organisation)
+
+                if not entity:
+                    # no entity for this organisation: log an unknown-entity
+                    # issue and emit no row for it
+                    self._log_unknown_entity(reference, curie, line_number)
+                    continue
+
+                entity = self.redirect_entity(entity)
+                if not entity or entity in emitted_entities:
+                    continue
+                emitted_entities.add(entity)
+
+                new_block = dict(block)
+                new_block["row"] = dict(row)
+                new_block["row"][self.entity_field] = entity
+                new_block["row"]["organisation"] = organisation
+                new_block["organisation"] = organisation
+                if self.issues:
+                    self.issues.record_entity_map(block["entry-number"], entity)
+                yield new_block
 
 
 class FactLookupPhase(LookupPhase):
--- a/digital_land/phase/lookup.py
+++ b/digital_land/phase/lookup.py
@@ -297,9 +297,9 @@ class EntityLookupPhase(LookupPhase):
                 yield block
                 continue
 
-            yield from self._fan_out(block)
+            yield from self._get_provider_entity_numbers(block)
 
-    def _fan_out(self, block):
+    def _get_provider_entity_numbers(self, block):
         """Resolve the entity for a row that has no organisation of its own.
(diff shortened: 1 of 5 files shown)


digital_land/phase/lookup.py
```diff
@@ -189,14 +214,82 @@ def process(self, stream):
...
+        # Single/zero-organisation resources are unchanged (byte-identical).
+        if len(self.organisations) > 1:
+            yield from self._process_multi_org(stream)
+            return
+
         for block in super().process(stream):
             if self.issues:
                 self.issues.record_entity_map(
                     block["entry-number"], block["row"]["entity"]
                 )
```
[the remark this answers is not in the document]
tombrooks248: No possible for this value to be falsy due to earlier running code - dismissing.

digital_land/phase/lookup.py
```diff
@@ -189,14 +214,85 @@ def process(self, stream):
...
+        super().__init__(
+            lookups=lookups,
+            redirect_lookups=redirect_lookups,
+            issue_log=issue_log,
+            operational_issue_log=operational_issue_log,
+            entity_range=entity_range,
+            lookup_rules=lookup_rules,
+        )
+        self.organisations = organisations or []
+
```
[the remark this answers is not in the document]
tombrooks248: This is constent with current codebase and in fact original code this is amending - ignorning.

digital_land/phase/lookup.py
```diff
@@ -84,14 +84,14 @@ def check_associated_organisation(self, entity):
                     return ""
         return entity
 
-    def get_entity(self, block):
+    def get_entity(self, block, organisation=None):
```
eveleighoj, line 87: add a docstring mainly to explain why we need an organisation arguement

digital_land/phase/lookup.py
```diff
@@ -189,14 +214,85 @@ def process(self, stream):
...
 
+    def __init__(
+        self,
+        lookups={},
+        redirect_lookups={},
+        issue_log=None,
+        operational_issue_log=None,
+        entity_range=[],
+        lookup_rules=None,
+        organisations=None,
```
eveleighoj: shall we call this providers instead of organisations to start being specific about what it is?

similar to the priority phase here https://github.com/digital-land/digital-land-python/blob/fa6580adc9d1b07396d44cc8f7d36f573c79ca3e/digital_land/phase/priority.py#L12

digital_land/phase/lookup.py
```diff
@@ -189,14 +214,85 @@ def process(self, stream):
...
+            if not prefix or row.get(self.entity_field, ""):
+                if self.issues:
+                    self.issues.record_entity_map(
+                        block["entry-number"], row.get(self.entity_field, "")
+                    )
+                yield block
+                continue
+
+            emitted_entities = set()
+            for organisation in self.organisations:
```
eveleighoj: Does this stop reading the organisation from the block and only use the providers?  so if the org on the block is different to the providers (which it can be mainly for some historic brownfield-land stuff) than the entity may not get assigned. (I might be missing something)

digital_land/phase/lookup.py
```diff
@@ -189,14 +214,85 @@ def process(self, stream):
...
+
         for block in super().process(stream):
             if self.issues:
                 self.issues.record_entity_map(
                     block["entry-number"], block["row"]["entity"]
                 )
             yield block
 
+    def _process_multi_org(self, stream):
+        for block in stream:
```
eveleighoj, line 287: It feels like some of this could stay in the higher level process function rather than all of it be separated into a function

digital_land/phase/lookup.py
```diff
@@ -189,14 +214,85 @@ def process(self, stream):
...
+                        block["entry-number"], row.get(self.entity_field, "")
+                    )
+                yield block
+                continue
+
+            emitted_entities = set()
+            for organisation in self.organisations:
+                entity = self.get_entity(block, organisation=organisation)
+
+                if not entity:
```
eveleighoj: I think we discussed this but how will we know which organisation there isn't an assigned entity for? by the time it gets to tasks it'll probably be for all organisations and will be confusing as the entity will be getting assigned. I would assume we only raise an issue if for no organisations (whether the org on the row or the providers) we cannot not assign it to an entity number.

eveleighoj (changes requested):

digital_land/phase/lookup.py
tombrooks248: Sure done.

digital_land/phase/lookup.py
tombrooks248: Yea good idea, and it fits with the way its done in the code base - changed.

digital_land/phase/lookup.py
tombrooks248: Sure I've had a go at that - is this better?

digital_land/phase/lookup.py
tombrooks248: Yep you're right. 

Five collections map a source column to organisation (brownfield-site's OrganisationURI, borough on conservation-area and central-activities-zone, plus developer-contributions and listed-building), so a row can carry an org that isn't in the providers list, and the fan-out func  currently ignores it. Caude tells me this wouldn’t be a problem with the data as it currently is but seems worth adding a guard incase data like this appears in the future.  

Added some code so if the row already knows its organisation, trust that and resolve with it; only try every provider when the row has nothing to go on.

digital_land/phase/lookup.py
tombrooks248: We won't know 

Ah yea I wrote this up in the PR, I think I erred on the side of false positives rather than false negatives, but you are right false positives are worse than false negatives. As they will be really annoying for for LPAs to be told their data (and quite a lot of data) is missing when its actually not missing anymore its just provided to us in a slightly odd way. 

I've made a change so we only raise an issue if we cannot assign an entity number to any organisation.

eveleighoj (dismissed):

eveleighoj (approved):

PR #578, opened 2026-07-21
switch Nan values to None

### LINKED ISSUE #2826: BUG: add_data fails with Postgrest JSONB NaN is invalid - multiple organisation change
An `add_data` request for the `developer-agreement` dataset fails when saving the response. The transform completes, but `save_response_to_db` → `_save_response_details` → `session.commit()` raises:

```
sqlalchemy.exc.DataError: (psycopg2.errors.InvalidTextRepresentation)
invalid input syntax for type json
DETAIL: Token "NaN" is invalid.
CONTEXT: JSON data, line 1: ...national-park-authority:Q72617160", "entity": NaN...
[SQL: INSERT INTO response_details (response_id, detail) ...]
```

A Python `float('nan')` reaches a `JSONB` column. `json.dumps` emits a bare `NaN` (it allows non-finite floats by default); Postgres `JSONB` follows strict JSON and rejects it, so the whole batch insert fails.

## Reproduction

- **Request:** `type: add_data`, `collection: developer-contributions`, `dataset: developer-agreement`, New Forest NPA source (`Developer-Agreement-2023.xlsx`), resolved `organisations = ['national-park-authority:Q72617158']`.
- **Fails** with digital-land-python `main` (`0.1.1.dev210+g9240853a7`).
- **Passes** with digital-land-python branch `109-bug-refresh-async-files-if-spec-has-updated`.

## Confirmed during debugging

- It takes the **single-organisation** path: `organisations` has length 1, so `EntityLookupPhase` uses `super().process()`, **not** the new multi-org fan-out.
- **Entity lookup works** — rows resolve normally (e.g. reference `09/93881-da` → entity `4601954`).
- **The issue log is clean** — after `IssueLog.add_severity_column`, `self.rows` contains **no** NaN (checked with `[r for r in self.rows if any(v!=v for v in r.values())]` → `[]`). So `pipeline-issues` is **not** the source, despite the error text mentioning `entity`.
- Therefore the NaN is elsewhere in the response detail (`converted_row` / `transformed_row` / a pipeline-summary breakdown). **Exact location not yet pinned**

## Leading hypothesis (not yet proven)

A change on digital-land-python `main` that is absent from branch `109-...` introduced this. Strongest suspects, both touching entity/organisation resolution.

Suspected interaction: a row that has an `organisation` value **in a column** plus a resource default organisation ends up producing a non-finite/blank `entity` that is serialised as a float `NaN` rather than `""` / `null`.

## Two separate problems

1. **Robustness (the crash):** an unresolved/blank entity must never serialise as `NaN`. Coerce non-finite floats → `None` before the `JSONB` write — either in digital-land-python where the value is produced, or defensively at the async-backend boundary in `_save_response_details`. This turns a hard `FAILED` into a gracefully-reported row.
2. **Behavioural (the trigger):** identify the exact `main` change producing the `NaN` for the org-column-plus-default case, and fix the resolution so the value is valid.

pooleycodes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When adding the severity column in `log.py`, Pandas represents missing values as `NaN`. Serialising these to JSON produces `NaN` tokens, which PostgREST rejects for a `jsonb` column ("NaN is invalid"), causing `add_data` to fail. This converts Pandas `NaN` values to `None` so they serialise as `null`/blank.

Can see  in [digital_land/check.py‎](https://github.com/digital-land/digital-land-python/pull/577/changes#diff-f09ae883d2744ef60541bf49e9becb809e91e555bfb39bb113a3d0234e5e0d54), entity is now added to the issue log, this will be added as a NaN value here, (is it worth adding to the issue log here?) and this is NaN causes json write issues.

- Ticket Link
- Related Issue #
- Closes #

images for UI changes._

## Added/updated tests?

- [ ] Yes
      have not been included_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing values in issue log records by consistently normalising merged record fields so that missing data is represented as blank/null values rather than `NaN`.

### DIFF
--- a/digital_land/log.py
+++ b/digital_land/log.py
@@ -167,10 +167,7 @@ class IssueLog(Log):
             self.fieldnames.append("responsibility")
             # Switch Pandas NaN to None to avoid issues with JSON
             self.rows = [
-                {
-                    k: (None if isinstance(v, float) and v != v else v)
-                    for k, v in row.items()
-                }
+                {k: (None if pd.isna(v) else v) for k, v in row.items()}
                 for row in merged_df.to_dict(orient="records")
             ]
 
--- a/digital_land/log.py
+++ b/digital_land/log.py
@@ -165,7 +165,14 @@ class IssueLog(Log):
             self.fieldnames.append("severity")
             self.fieldnames.append("description")
             self.fieldnames.append("responsibility")
-            self.rows = merged_df.to_dict(orient="records")
+            # Switch Pandas NaN to None to avoid issues with JSON
+            self.rows = [
+                {
+                    k: (None if isinstance(v, float) and v != v else v)
+                    for k, v in row.items()
+                }
+                for row in merged_df.to_dict(orient="records")
+            ]
 
     def appendErrorMessage(self, mapping_path):
         # Read the mapping from the JSON config file


gibahjoe (approved):
LGTM

PR #579, opened 2026-08-10
conditional fields expectation for local plans

tombrooks248:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds a new dataset expectation, `check_fields_required_after_plan_event`, which checks that a three specific fields are populated on plan records — but only for plans that have actually reached the stage where those fields become required. 

A plan reaches that stage when its linked `plan-timetable` row for `proposed-plan-consultation-start` has an `actual-date` in the past. 

**This change is inert on merge** — the operation only runs once a matching rule is added to `expect.csv`, which is a later step. 

## Why

We used to flag these fields as "missing value" whenever they were blank, producing false positives on the LPA Dashboard — LPAs were asked for data they weren't yet expected to supply. Part 1 of the ticket removed those checks, which stopped the false positives but left no check at all. This restores it, gated on the plan's actual progress.

It's an expectation rather than an issue because the gate lives in a different dataset from the fields being checked. Issues are raised per-resource during processing and can't see across datasets; expectations run after the data is built, so the comparison is straightforward.

 Only one plan has passed `proposed-plan-consultation-start` with a recorded `actual-date`, and it's fully populated — most LPAs have that consultation scheduled for 2027-28. So this is really work that all about getting our systems ready for data that we expect to exist in the future.

- [Ticket Link](https://github.com/digital-land/submit/issues/1187)
- Related Issue #
- Closes #

This change is inert on merge: the new operation is only ever reached by name from a rule in `expect.csv`, and no rule references it yet — across all 55 `expect.csv` files in config, only `count_deleted_entities`, `count_lpa_boundary` and `duplicate_geometry_check` are used.

Nothing will run, and no builds change behaviour. 

After this has been merged I will make a PR for config repo and for collection-task do some testing from there in DEV.

## Added/updated tests?

- [x] Yes
      have not been included_

### DIFF
--- a/digital_land/expectations/checkpoints/dataset.py
+++ b/digital_land/expectations/checkpoints/dataset.py
@@ -12,6 +12,7 @@ from digital_land.organisation import Organisation
 from .base import BaseCheckpoint
 from ..log import ExpectationLog
 from ..operations.dataset import (
+    check_fields_required_after_plan_event,
     count_lpa_boundary,
     count_deleted_entities,
     duplicate_geometry_check,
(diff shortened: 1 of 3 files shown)


gibahjoe (approved):
LGTM

PR #582, opened 2026-08-20
feat: add name_is_a_code expectation

tombrooks248:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds a new dataset expectation, `name_is_a_code_check`, which flags entities whose `name` is only a bare reference code rather than a description — for example `59`, `3B` or `11/802` — and registers it in the dataset checkpoint's operation map. Failures are reported per entity with their `organisation_entity`, so they can be attributed back to the provider.

The check is inert until it is switched on for a dataset by an `expect.csv` row in `config`, so this PR changes no behaviour on its own. It is intended for `article-4-direction-area`, `conservation-area` and `listed-building-outline` only.

## Why

Part of the work to give LPAs feedback on `name` field quality. Analysis of the ODP spatial datasets found a significant number of entities whose name carries no information for a data user, because the provider has supplied their internal reference instead of a descriptive name.

Note that nothing is yet visible to an LPA: expectations do not currently generate tasks. That bridge is config#2912 and is deliberately out of scope here.

- [Ticket Link](https://github.com/digital-land/config/issues/2909)
- Related Issue #
- Closes #

The check was also run against the current published data for all three datasets in scope, and the counts reconcile exactly with the 2026-08 ODP name analysis:

| dataset | flagged | analysis figure |
|---|---|---|
| article-4-direction-area | 104 of 7,327 (1.4%) | 104 (1.4%) |
| conservation-area | 10 of 11,019 (0.1%) | 10 (0.1%) |
| listed-building-outline | 144 of 126,271 (0.1%) | 144 (0.1%) |

That is 258 entities across roughly 20 organisations in total.

## Context - low risk change:

The operation does nothing until an `expect.csv` row in config names it, so merging this changes no behaviour anywhere. When it is switched on it runs at `warning` severity, which cannot fail a dataset build, and it only reads the `entity` table — it writes nothing back to the dataset. Nothing is visible to a data provider yet either, because expectations do not currently generate tasks.

## Future testing before this goes live.

Once this is merged I will set up the PR to change in config and then test that in DEV before merging to main.

## Added/updated tests?

- [x] Yes
      have not been included_

### DIFF
--- a/digital_land/expectations/checkpoints/dataset.py
+++ b/digital_land/expectations/checkpoints/dataset.py
@@ -17,6 +17,7 @@ from ..operations.dataset import (
     count_deleted_entities,
     duplicate_geometry_check,
     fetch_active_resources_for_dataset,
+    name_is_a_code_check,
 )
 
 
@@ -42,6 +43,7 @@ class DatasetCheckpoint(BaseCheckpoint):
             "count_lpa_boundary": count_lpa_boundary,
             "count_deleted_entities": count_deleted_entities,
             "duplicate_geometry_check": duplicate_geometry_check,
+            "name_is_a_code_check": name_is_a_code_check,
         }
         operation = operation_map[operation_string]
         return operation
--- a/digital_land/expectations/operations/dataset.py
+++ b/digital_land/expectations/operations/dataset.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import re
 import sqlite3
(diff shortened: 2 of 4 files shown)


pooleycodes (approved):
LGTM

PR #585, opened 2026-08-24
feat: add duplicate_name_check expectation

tombrooks248:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## What

Adds a `duplicate_name_check` dataset expectation, which flags names reused by several entities within the same provision — the same dataset and organisation.

## Why

Part 3 of digital-land#598. The same `name` text is reused across distinct entities, so the name stops identifying anything — usually one feature split into several polygons whose parts were never distinguished. For example one authority used `"Conservation_Area.TAB"` for 27 conservation areas, and another used `"District Centre related A4D"` for 364 article-4-direction areas.

Grouping is per organisation rather than dataset-wide because the published dataset is every LPA's file concatenated together; `"City Centre"` appears once each in seven different councils, and none of them can act on that. Matching reuses the existing `_normalise_name` helper, so case and spacing variants group together while the raw spelling is what gets reported.

- [Ticket Link](https://github.com/digital-land/config/issues/2911)
- Related Issue #
- Closes #

## Testing

5 integration tests covering per-organisation scoping, the threshold parameter, normalisation, blank exclusion and the failure structure. 105 expectations tests pass, black and flake8 clean.

Checked against currently published data: 466 duplicate-name groups for `article-4-direction-area` and 178 for `conservation-area`.

## This change is inert
This change is inert on its own. The operation only runs for a dataset once a row naming it is added to that collection's `expect.csv`, so merging this alters no pipeline output. Opting in `article-4-direction-area` and `conservation-area` follows in a separate config PR.

## Added/updated tests?

- [x] Yes
      have not been included_

## Summary by CodeRabbit

- **New Features**
  - Added dataset validation to detect repeated entity names.
  - Matching is case- and whitespace-insensitive, while blank names are excluded.
  - Supports configurable thresholds for identifying duplicates.
  - Results are consistently ordered and include relevant entity references.

- **Tests**
  - Added coverage for duplicate detection, dataset and organisation scoping, name normalisation, blank-name handling, and configurable thresholds.

### DIFF
--- a/digital_land/expectations/operations/dataset.py
+++ b/digital_land/expectations/operations/dataset.py
@@ -641,6 +641,83 @@ def name_is_a_placeholder_check(conn, listed_building_path, placeholders=None):
     return result, message, details
 
 
+def duplicate_name_check(conn, threshold=2):
+    """
+    Checks for names reused across several entities within the same provision, i.e.
+    the same dataset and organisation. Usually means one feature was split into
+    several polygons and the parts were never distinguished.
+
+    Grouping is per organisation, not dataset-wide. The published dataset is every
+    LPA's file concatenated together, so a dataset-wide comparison flags authorities
+    for other authorities' word choices - 'City Centre' appears once each in seven
+    different councils, and none of them can act on that.
+
+    Failures are one entry per duplicated name rather than per entity, matching the
+    analysis this came from. Member entity ids are nested under 'entities' so a
+    drill-down page can page through them without re-running the check.
+
+    Matching is case- and whitespace-insensitive via _normalise_name, but not
+    otherwise fuzzy: two directions differing only by year are different names.
+
+    Blank names are not compared - they are already covered by the missing values
+    issue, and every blank would otherwise be a duplicate of every other blank.
+
+    args:
+        conn: connection to the dataset being checked, created by the checkpoint class
+        threshold: how many entities must share a name before it is reported.
+            Exposed so it can be raised from config without a code release.
+    """
+    query = """
+        select entity, reference, name, organisation_entity
+        from entity
+        where name is not null
+            and trim(name) != ''
+    """
+    rows = conn.execute(query).fetchall()
+
+    groups = {}
+    for entity, reference, name, organisation_entity in rows:
+        key = (organisation_entity, _normalise_name(name))
+        # keep the first raw spelling seen: the normalised key is not worth showing to
+        # an LPA, and two rows differing only in case would otherwise look identical
+        group = groups.setdefault(key, {"name": name, "entities": [], "references": []})
+        group["entities"].append(entity)
+        group["references"].append(reference)
+
+    failures = [
+        {
+            "organisation_entity": organisation_entity,
+            "name": group["name"],
+            "count": len(group["entities"]),
+            "entities": sorted(group["entities"]),
+            "references": sorted(
+                reference for reference in group["references"] if reference
+            ),
+        }
+        for (organisation_entity, _), group in groups.items()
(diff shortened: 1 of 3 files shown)


digital_land/expectations/operations/dataset.py
```diff
@@ -641,6 +641,83 @@ def name_is_a_placeholder_check(conn, listed_building_path, placeholders=None):
...
+            and trim(name) != ''
+    """
+    rows = conn.execute(query).fetchall()
+
+    groups = {}
+    for entity, reference, name, organisation_entity in rows:
+        key = (organisation_entity, _normalise_name(name))
+        # keep the first raw spelling seen: the normalised key is not worth showing to
+        # an LPA, and two rows differing only in case would otherwise look identical
+        group = groups.setdefault(key, {"name": name, "entities": [], "references": []})
```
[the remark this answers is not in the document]
tombrooks248: Good idea - done

digital_land/expectations/operations/dataset.py
```diff
@@ -641,6 +641,90 @@ def name_is_a_placeholder_check(conn, listed_building_path, placeholders=None):
     return result, message, details
 
 
+def duplicate_name_check(conn, threshold=2):
+    """
```
gibahjoe, line 645: The function doc says it checks within a provision however, the connection could contain data from different datasets (its a public function) and the grouping process does not  group by dataset. 

This means you could run this check  across different provisions and end up with duplicates.

tests/integration/expectations/operations/test_dataset.py
```diff
@@ -968,3 +969,118 @@ def test_name_is_a_placeholder_check_single_string_placeholder(
...
+        # same name twice in one org, should be flagged
+        (1, "A4D-1", "Site allocation NSP30", "600001"),
+        (2, "A4D-2", "Site allocation NSP30", "600001"),
+        # three in one org, should be flagged with a count of 3
+        (3, "A4D-3", "District Centre related A4D", "600001"),
+        (4, "A4D-4", "District Centre related A4D", "600001"),
+        (5, "A4D-5", "District Centre related A4D", "600001"),
+        # used once, should not be flagged
+        (6, "A4D-6", "Queen's Park", "600001"),
+    ]
```
gibahjoe, line 993: Add a test case where duplicate names exist across different datasets, to ensure the behaviour remains consistent.

gibahjoe (dismissed):
More of a nitpick. 

Also consider running the grouping in the connection since you already have it which will probably be beneficial for humongous rows.

digital_land/expectations/operations/dataset.py
tombrooks248: Good idea - yea I have fixed the code to make it do what it says it does, and hopefully make it more useful if someone else wants to re-use the code in future.

tests/integration/expectations/operations/test_dataset.py
tombrooks248: Yes and added a relevant test for it.

tombrooks248:
> More of a nitpick.
> 
> Also consider running the grouping in the connection since you already have it which will probably be beneficial for humongous rows.

On this - I looked into it and because its running the python function `_normalise_name(name)` which does some things in python (regex, internal trimming, handling tabs and newlines) which would be be very complicated/ possible not doable in SQL, I don't think its worth refactoring into SQL. But I agree in general doing this stuff in SQL is better when that is possible.

gibahjoe (approved):
LGTM

PR #584, opened 2026-08-24
fix: update PrintLookupPhase to allow processing without organisation

gibahjoe:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Allow `PrintLookupPhase` to create a lookup entry when an organisation is not provided, as long as a prefix and reference are present.

## Related Tickets & Documents

- Ticket Link: N/A
- Related Issue #: N/A
- Closes #: N/A

## QA Instructions, Screenshots, Recordings

Run:

```sh
.venv/bin/pytest tests/unit/phase/test_lookup.py
.venv/bin/black --check digital_land/phase/lookup.py tests/unit/phase/test_lookup.py
.venv/bin/flake8 digital_land/phase/lookup.py tests/unit/phase/test_lookup.py
```

All checks pass. No UI changes.

## Added/updated tests?

- [x] Yes
- [ ] No, and this is why:
- [ ] I need help with writing tests

## [optional] Are there any post deployment tasks we need to perform?

None.

## [optional] Are there any dependencies on other PRs or Work?

None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Lookup entries are now created when a prefix and reference are provided, even if the organisation field is blank.
  * Blank organisation details are retained in the resulting lookup entry.

PR #586, opened 2026-08-24
fix: allow lookup entries without an organisation to validate

pooleycodes:
PR #584 relaxed `PrintLookupPhase` so a lookup entry is emitted for rows with no organisation:

```diff
-                if prefix and organisation and reference:
+                if prefix and reference:
```

but `Lookups.validate_entry` still requires one:

```python
expected_fields = ["prefix", "organisation", "reference"]
```

so digital-land now produces lookup entries that digital-land itself rejects with:

```
ValueError: ERROR: expected organisation not found in lookup entry
```

## Impact

Any consumer that feeds `PrintLookupPhase` output into `Lookups.add_entry` hits this. In `async-request-backend/request-processor` it surfaces as two failing integration tests, where the whole workflow returns a 500 instead of a result:

```
FAILED tests/integration/src/application/core/test_workflow.py::test_run_workflow
FAILED tests/integration/src/application/core/test_workflow.py::test_run_workflow_brownfield_land
AssertionError: assert 'converted-csv' in {'exception': 'ValueError', 'status': 500, ...}
```

Verified against request-processor with only the digital-land commit varying:

| digital-land | result |
| --- | --- |
| `fac1d93` (current main) | 2 failed, 2 passed |
| `fac1d93` + this change | **4 passed** |

## Why CI did not catch it

The two halves are each covered, but nothing joins them up:

- the test added in #584 asserts only what `PrintLookupPhase` *emits*, and never passes that entry to `validate_entry`
- every case in `test_lookups_validate_entry_failure` has an empty `prefix`, so they all raise on `prefix` and never reach the `organisation` check

This PR adds a regression test for the missing combination — a complete entry with a blank `organisation`.

## For reviewer attention

Removing the guard means blank-organisation entries now reach the duplicate check below it:

```python
if item["prefix"] == entry["prefix"]
and item["organisation"] == entry["organisation"]
and item["reference"] == entry["reference"]
```

Two entries sharing `prefix` + `reference` with blank organisations will now dedup as the same entry, where previously neither reached this code. If blank-organisation lookups can represent genuinely different organisations, that would collapse them into one entity — worth a check from whoever owns #584's intent.

## Testing

- `tests/unit/test_pipeline.py` and `tests/unit/phase/test_lookup.py` lookup tests: 9 passed (8 existing + 1 new)
- no change to the rest of the suite: 59 passed both before and after
- `black` and `flake8` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Lookup entries can now be accepted without an organisation, provided the prefix and reference fields are completed.
  * Duplicate detection continues to distinguish entries using the organisation, prefix and reference.

### DIFF
--- a/digital_land/pipeline/main.py
+++ b/digital_land/pipeline/main.py
@@ -943,7 +943,7 @@ class Lookups:
     # @staticmethod
     def validate_entry(self, entry) -> bool:
         # ensures minimum expected fields exist and are not empty strings
-        expected_fields = ["prefix", "organisation", "reference"]
+        expected_fields = ["prefix", "reference"]
         for field in expected_fields:
             if not entry.get(field, ""):
                 raise ValueError(f"ERROR: expected {field} not found in lookup entry")
--- a/digital_land/pipeline/main.py
+++ b/digital_land/pipeline/main.py
@@ -955,7 +955,7 @@ class Lookups:
                     1
                     for item in self.entries
                     if item["prefix"] == entry["prefix"]
-                    and item["organisation"] == entry["organisation"]
+                    and item.get("organisation", "") == entry.get("organisation", "")
                     and item["reference"] == entry["reference"]
                 ]
             )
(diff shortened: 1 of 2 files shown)


gibahjoe (approved):
LGTM

PR #590, opened 2026-09-02
fix: entity priority - order on resource_start_date before entry_number

tombrooks248:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Reorders the entity value-selection `ORDER BY` in `digital_land/package/dataset_parquet.py` so resource recency outranks row position.

```sql
-- before
ORDER BY priority DESC, entry_date DESC, entry_number DESC, resource_end_date DESC, tf.resource, fact
-- after
ORDER BY priority DESC, entry_date DESC, resource_end_date DESC, resource_start_date DESC, entry_number DESC, tf.resource, fact
```

`entry_number` becomes the final tie-breaker it was always meant to be. Adds four integration tests, one pinned to each criterion, and a `start-date` column to the resource fixtures.

## Why

config#2945 (DLSD-3458): a Leicester listed-building-outline entity continued publishing its April 2025 geometry after the authority corrected it in August 2026. Both values are present in the fact table and both carry the same `entry_date` — the listed building's own 1975 designation date rather than a collection date — so the tie fell straight through to `entry_number`, letting row position within a file decide the published value instead of which resource was more recent.

Both resource dates are needed. `resource_end_date` coalesces a live resource to `2999-12-31`, so it correctly ranks live resources above superseded ones and orders the rest by recency — but it cannot separate two live resources, which all land on that same `2999-12-31`. `resource_start_date` breaks those, and is what the ticket's third criterion asks for: measured across eight collections it uniquely resolves 1,260 of the 5,261 value changes.

The existing tests could not have caught any of this. So I have added some new tests to catch this going forwards.

- [Ticket Link](https://github.com/digital-land/config/issues/2945)
- Related Issue #
- Closes #

I have both new and old version of the code locally but with the published `fact`/`fact_resource` tables and `collection/resource.csv` for eight collections (54,148 entities): infrastructure-project, green-belt, special-area-of-conservation, agricultural-land-classification, local-plan, article-4-direction, conservation-area and brownfield-land. 5,261 values change and **every one moves from a superseded resource to a more recent one** — no case anywhere selects an older resource.

The only user-visible consequence is in brownfield-land: **170 sites across 28 LPAs gain an end-date and leave the active register**, and 5 move the other way and become active again. 151 of the 170 take their new end-date from the authority's currently live resource, so this corrects sites the old ordering was holding open against the LPA's own published data. No other collection tested produces an active-to-ended transition.

Adding `resource_start_date` accounts for 1,260 of the 5,261 changes — cases where competing values all come from currently live resources, so `resource_end_date` ties at the sentinel and the decision previously fell back to `entry_number`.

### Tested in DEV

I have pushed this change to DEV and run article-4-direction - all tasks ran green.

## Added/updated tests?

- [x] Yes
      have not been included_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved entity value selection so active and later-ending resources take precedence over earlier-ended resources.
  * When multiple active resources are available, the later-starting resource is selected.
  * Preserved entry-number ordering when records belong to the same resource.

* **Tests**
  * Added regression coverage for resource date precedence and entry-number tie-breaking.

### DIFF
--- a/digital_land/package/dataset_parquet.py
+++ b/digital_land/package/dataset_parquet.py
@@ -600,15 +600,14 @@ class DatasetParquetPackage(Package):
             SELECT {fields_str}{optional_org_str}, priority FROM (
                 SELECT {fields_str},
                 MAX(priority) OVER (PARTITION BY entity) AS priority,
-                CASE WHEN resource_csv."end-date" IS NULL THEN '2999-12-31' ELSE resource_csv."end-date" END AS resource_end_date,
-                CASE WHEN resource_csv."start-date" IS NULL THEN '0001-01-01' ELSE resource_csv."start-date" END AS resource_start_date
+                CASE WHEN resource_csv."end-date" IS NULL THEN '2999-12-31' ELSE resource_csv."end-date" END AS resource_end_date
                 FROM parquet_scan('{transformed_parquet_dir}/{parquet_str}') tf
                 LEFT JOIN read_csv_auto('{resource_path}', max_line_size=40000000) resource_csv
                 ON tf.resource = resource_csv.resource
                 {entity_where_clause}
                 QUALIFY ROW_NUMBER() OVER (
                     PARTITION BY entity, field
-                    ORDER BY priority DESC, entry_date DESC, resource_end_date DESC, resource_start_date DESC, entry_number DESC, tf.resource, fact
+                    ORDER BY priority DESC, entry_date DESC, resource_end_date DESC, resource_csv."start-date" DESC, entry_number DESC, tf.resource, fact
                 ) = 1
             )
         """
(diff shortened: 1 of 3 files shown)


digital_land/package/dataset_parquet.py
```diff
@@ -600,14 +600,15 @@ def load_entities_range(
             SELECT {fields_str}{optional_org_str}, priority FROM (
                 SELECT {fields_str},
                 MAX(priority) OVER (PARTITION BY entity) AS priority,
-                CASE WHEN resource_csv."end-date" IS NULL THEN '2999-12-31' ELSE resource_csv."end-date" END AS resource_end_date
+                CASE WHEN resource_csv."end-date" IS NULL THEN '2999-12-31' ELSE resource_csv."end-date" END AS resource_end_date,
+                CASE WHEN resource_csv."start-date" IS NULL THEN '0001-01-01' ELSE resource_csv."start-date" END AS resource_start_date
```
Ben-Hodgkiss: Are there any resources without a start date? Surely not?

tombrooks248: Good point, removed this change.

jpherr (approved):
👍

PR #594, opened 2026-09-04
fix: prevent tests breaking on fact_resrouce col change

tombrooks248:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Replaces the hardcoded column-count assertions in `test_dataset_parquet.py` with assertions that derive the expected columns from the specification, and checks column names and order rather than just the count.

## Why

Tests checking the table shape in specification repo should be in the specification repo not the digital-land-python repo

These tests hardcoded `fact.parquet` as 9 columns and `fact-resource.parquet` as 7. The specification is fetched from the specification repo's `main` branch at build time, so adding `entity` and `field` to `fact-resource` would have broken CI here the moment that change merged, and removing those fields from `fact` later would break it again. Deriving the expectation from the spec removes that coupling, and asserting on names and order is strictly stronger than a count — it catches a renamed or reordered column, which matters because the sqlite insert in `dataset_parquet.py` is positional.

- [Ticket Link](https://github.com/digital-land/config/issues/2960)
- Related Issue #
- Closes #

This is just an edit to tests to tests passing is enough.

## Added/updated tests?

- [x] Yes
      have not been included_

## Tests

* Updated integration checks to validate exported Parquet columns against the package schema rather than fixed column counts.
* Improved coverage for fact and fact-resource outputs, including schema-defined fields and field ordering.
* Expanded SQLite test fixtures to include entity and field data, with support for dynamically generated fact-resource columns.
* Continued coverage for single-file, multi-file, batch, and empty-file scenarios.

### DIFF
--- a/tests/integration/package/test_dataset_parquet.py
+++ b/tests/integration/package/test_dataset_parquet.py
@@ -397,7 +397,13 @@ def test_load_facts_single_file(data: dict, expected: int, tmp_path):
     assert (
         len(df) == expected
     ), "No. of facts does not match expected"  # No of unique facts
-    assert df.shape[1] == 9, "Not all columns saved in fact.parquet file"
+    expected_columns = [
+        field.replace("-", "_")
+        for field in package.specification.schema["fact"]["fields"]
+    ]
+    assert (
+        list(df.columns) == expected_columns
+    ), "fact.parquet columns do not match the specification"
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


tests/integration/package/test_dataset_parquet.py
```diff
@@ -397,7 +397,13 @@ def test_load_facts_single_file(data: dict, expected: int, tmp_path):
     assert (
         len(df) == expected
     ), "No. of facts does not match expected"  # No of unique facts
-    assert df.shape[1] == 9, "Not all columns saved in fact.parquet file"
+    expected_columns = [
+        field.replace("-", "_")
+        for field in package.specification.schema["fact"]["fields"]
+    ]
```
jpherr, line 414: Does this mean if the spec was updated, it would break the pipeline in this app?

tombrooks248: So currently yes if the spec was updated it would break the pipeline, but after this change it should be possible to update the specifiscation repo for this table and it not break these pipelines.

jpherr (approved):
👍

jpherr:
Interesting that approvals are retained even after new changes are pushed. Is that desired behaviour?

PR #595, opened 2026-09-07
fix: add fact resrouce columns to test fexiture

tombrooks248:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds `entity` and `field` columns to the `fact_resource` table in the committed test fixture `tests/data/dataset/central-activities-zone.sqlite3`.

## Why

The specification is about to gain those two fields on `fact-resource`. `dataset-update` builds its INSERT column list from the specification, so once that lands it inserts nine named columns into this fixture's seven-column table and fails with `sqlite3.OperationalError: table fact_resource has no column named entity`. That breaks six tests across `tests/acceptance/test_add_data.py` and `tests/acceptance/test_commands.py`, which both use this fixture.

## What the binary diff actually contains

GitHub can't render this, so for review — the only change is two columns appended to one table definition:

```diff
- start_date TEXT, FOREIGN KEY (fact) REFERENCES fact (fact))
+ start_date TEXT, entity INTEGER, field TEXT, FOREIGN KEY (fact) REFERENCES fact (fact))
```

Produced by `ALTER TABLE fact_resource ADD COLUMN entity INTEGER` and `ADD COLUMN field TEXT`. The file is the same size before and after (253,952 bytes) because SQLite only rewrites the schema record; no data pages change.

No rows were changed. The columns are appended, and the insert is by column name rather than position, so a wider table is backwards-compatible — the two columns simply take NULL until the specification declares them.

- [Ticket Link](https://github.com/digital-land/config/issues/2960)
- Related Issue #
- Closes #

## Testing

Ran all 142 test files individually against both the current seven-field specification and a locally patched nine-field one, and diffed the results: no test anywhere changes behaviour. Specifically `test_add_data.py` 7 passed and `test_commands.py` 2 passed under both, where previously they were 4 failed and 2 failed under the new specification. `tests/integration/test_add_data_utils.py`, the other consumer of this fixture, is unaffected at 25 passed.

## Added/updated tests?

- [x] Yes
      have not been included_


gibahjoe (approved):
LGTM

